PR #21, opened 2016-11-16
feat(component-list): Make component list page work with 2 views

tinayuangao:
R: @jelbourn  @mmalerba


src/app/pages/components/components.html
```diff
@@ -1,4 +1,28 @@
...
+<div class="docs-components-component-list">
+  <div class="docs-components-header-section">
+    <h1>Directory of Design Components</h1>
+    <p>
+      Set of high-quality UI components built with Angular 2 and
+      TypeScript, following the Material Design spec
+    </p>
+  </div>
+  <div [class.docs-components-grid-view]="gridMode" class="docs-components-view">
+    <button md-fab color="primary" (click)="switchViewMode()" class="docs-components-grid-fab">
```
jelbourn: This should actually be a button-toggle (I noticed this when we were looking at the mocks yesterday, there hasn't been time to update them yet)

src/app/pages/components/components.html
```diff
@@ -1,4 +1,28 @@
...
-  {{component.name}}
+<div class="docs-components-header-bar">Components</div>
+<div class="docs-components-component-list">
+  <div class="docs-components-header-section">
+    <h1>Directory of Design Components</h1>
+    <p>
+      Set of high-quality UI components built with Angular 2 and
+      TypeScript, following the Material Design spec
+    </p>
+  </div>
```
jelbourn: I'm wondering if this paragraph at the top is necessary. @amitafr thoughts?

src/app/pages/components/components.html
```diff
@@ -1,4 +1,28 @@
...
+  <div [class.docs-components-grid-view]="gridMode" class="docs-components-view">
+    <button md-fab color="primary" (click)="switchViewMode()" class="docs-components-grid-fab">
+      <md-icon>{{viewIcon}}</md-icon>
+    </button>
+    <div class="docs-components-list">
+      <div *ngFor="let group of componentGroups" class="docs-components-group">
+        <div class="docs-components-group-name">{{group.name}}</div>
+        <div class="docs-components-section">
```
jelbourn: Could we make this an `md-grid-list`?

src/app/pages/components/components.html
```diff
@@ -1,4 +1,28 @@
...
+      TypeScript, following the Material Design spec
+    </p>
+  </div>
+  <div [class.docs-components-grid-view]="gridMode" class="docs-components-view">
+    <button md-fab color="primary" (click)="switchViewMode()" class="docs-components-grid-fab">
+      <md-icon>{{viewIcon}}</md-icon>
+    </button>
+    <div class="docs-components-list">
+      <div *ngFor="let group of componentGroups" class="docs-components-group">
+        <div class="docs-components-group-name">{{group.name}}</div>
```
jelbourn: This should be a header. Probably `<h2>` (not sure 100% where we are in the hierarchy right now)

src/app/pages/components/components.html
```diff
@@ -1,4 +1,28 @@
...
+        <div class="docs-components-section">
+          <div *ngFor="let component of group.list">
+            <div class="docs-component-chip"  [routerLink]="component.routerLink" >
+              <div class="docs-component-img"><img [src]="component.imageSrc"/></div>
```
jelbourn: FYI: using `` is fine for now, but we'll eventually need to change these to be inline SVGs so we can style them based on the theme.

src/app/pages/components/components.ts
```diff
@@ -1,18 +1,64 @@
-import { Component} from '@angular/core';
+import {Component} from '@angular/core';
+
+export class ComponentGroup {
```
jelbourn: Add class description

src/app/pages/components/components.ts
```diff
@@ -1,18 +1,64 @@
...
 export class ComponentsList {
-  componentItems = [
-    {name: 'Buttons', src: 'button'},
-    {name: 'Cards', src: 'card'},
-    {name: 'Chips', src: 'chip'},
-    {name: 'Grid lists', src: 'grid'},
-    {name: 'Menu', src: 'menu'},
-    {name: 'Tooltip', src: 'tooltip'},
-    {name: 'Progress', src: 'progress'},
+  componentGroups: ComponentGroup[] = [
```
jelbourn: I think this would be a bit more readable if we defined the data like this:
```ts
componentGroups = groupComponents([
  {name: 'Buttons',  src: 'button',   group: 'Form controls'},
  {name: 'Checkbox', src: 'checkbox', group: 'Form controls'},
  ...
]);
```
Where `groupComponents` creates the data structures. Thoughts?

src/app/pages/components/components.html
amitafr: @jelbourn not necessary - can be removed

src/app/pages/components/components.ts
tinayuangao: Not sure it is more readable... The list and its hierarchy is relatively static.

PR #26, opened 2016-11-18
feat(plnkr): Add initial version of openPlunker

tinayuangao:
R: @jelbourn @mmalerba


src/app/shared/plnkr/plnkr-util.ts
```diff
@@ -0,0 +1,109 @@
...
+    `Copyright 2016 Google Inc. All Rights Reserved.
+    Use of this source code is governed by an MIT-style license that
+    can be found in the LICENSE file at http://angular.io/license`;
+
+  TEMPLATE_PATH = '/app/shared/plnkr/template/';
+  TEMPLATE_FILES = ['index.html', 'systemjs.config.js', 'main.ts'];
+
+  TAGS: string[] = ['angular', 'material', 'example'];
+
+  description: string = 'Example for Angular Material';
```
mmalerba: Maybe add all of these to some kind of ExampleDescriptor class?

src/app/shared/plnkr/plnkr.ts
```diff
@@ -0,0 +1,39 @@
+import { Component, Input, OnInit } from '@angular/core';
+import { Http } from '@angular/http';
+import { PlunkerUtil } from './plnkr-util';
+
+@Component({
+  selector: 'app-plnkr',
```
mmalerba: Maybe "plunker-example"? (also be consistent with plnkr vs plunker, my vote is for everything to be "plunker")

src/app/shared/plnkr/example/button-demo.html
```diff
@@ -0,0 +1,104 @@
+<div class="demo-button">
```
mmalerba: I know these will eventually go in a separate npm package, but maybe for now put them in assets/examples?

src/app/shared/plnkr/template/index.html
```diff
@@ -0,0 +1,36 @@
+<!DOCTYPE html>
```
mmalerba: can go in assets/plunker

src/app/shared/plnkr/plnkr-util.ts
```diff
@@ -0,0 +1,109 @@
...
+    input.value = value;
+    this.form.appendChild(input);
+  }
+
+  _readFile(filename: string, path: string) {
+    this._http.get(path + filename).subscribe(
+      (response) => {
+        let content = response.text();
+        content = this._processContent(filename, path, content);
+        this._generateInput(filename, content, true);
```
jelbourn: I'd break these out into separate handler functions:
```ts
 this._http.get(path + filename).subscribe(
  response => this._addFileToForm(response.text()),
  error => console.log(error));
```

Where `_addFileToForm` massages the file content and adds the input. Then you could remove the `isFile` argument from `_createFormInput` and pass the `files[${name}]` as the name (since you know specifically you're processing a file).

src/app/shared/plnkr/plnkr-util.ts
```diff
@@ -0,0 +1,109 @@
...
+      }, (error) => {
+        console.log(error);
+      }
+    );
+  }
+
+  _processContent(filename: string, path: string, content: string): string {
+    if (path == this.TEMPLATE_PATH) {
```
tinayuangao: Added comments

tinayuangao:
Added PlunkerData and changed PlunkerUtil to PlunkerWriter. PTAL. Thanks for review!

src/app/shared/plnkr/plunker-writer.ts
```diff
@@ -0,0 +1,128 @@
...
+ *   files[directory-name/file-name2]: file-content2,
+ *   // Can add multiple tags
+ *   tags[0]: tag-0,
+ *   // Description of plunker
+ *   description: description,
+ *   // Private or not
+ *   private: true
+ * }
+ */
+export class PlunkerWriter {
```
jelbourn: You can add the `@Injectable` decorator to this to turn it into a service and add it to the `providers` list of the app's `NgModule`. Then `PlunkerButton` can inject this (instead of http) and this service can inject `Http` directly.

src/app/shared/plnkr/plunker-writer.ts
```diff
@@ -0,0 +1,128 @@
...
+
+  COPYRIGHT =
+    `Copyright 2016 Google Inc. All Rights Reserved.
+    Use of this source code is governed by an MIT-style license that
+    can be found in the LICENSE file at http://angular.io/license`;
+
+  TEMPLATE_PATH = '/app/shared/plnkr/template/';
+  TEMPLATE_FILES = ['index.html', 'systemjs.config.js', 'main.ts'];
+
+  TAGS: string[] = ['angular', 'material', 'example'];
```
jelbourn: I think you missed moving these constants out of the `class`; they can go right above it ([example](https://github.com/angular/material2/blob/8caf9a6553d57f46ea88a4e306a5438bb2dac571/src/lib/slider/slider.ts#L31-L31))

src/app/shared/plnkr/plunker-writer.ts
```diff
@@ -0,0 +1,128 @@
...
+      error => console.log(error));
+  }
+
+  _addFileToForm(content: string, filename: string, path: string) {
+    this.counter++;
+    if (path == this.TEMPLATE_PATH) {
+      content = this._processTemplateFile(filename, path, content);
+    }
+    this._createFormInput(`files[${filename}]`, this._appendCopyright(filename, content));
+    this._submitForm();
```
jelbourn: Did you mean to move this `_submitForm` call to the end of `openPlunker`?

src/app/shared/plnkr/plunker-writer.ts
```diff
@@ -0,0 +1,128 @@
...
+  }
+
+  _processTemplateFile(filename: string, path: string, content: string): string {
+    if (filename == 'index.html') {
+      // Replace the component selector in `index,html`.
+      // For example, <material-docs-example></material-docs-exmaple> will be replaced as
+      // <button-demo></button-demo>
+      content = content.replace(new RegExp('material-docs-example', 'g'),
+        this.exampleData.selectorName);
```
jelbourn: You could use a regex literal:
```ts
content = content.replace(/material-docs-example/g, this.exampleData.selectorName),
```

src/app/shared/plnkr/plunker-writer.ts
```diff
@@ -0,0 +1,128 @@
...
+  _addFileToForm(content: string, filename: string, path: string) {
+    this.counter++;
+    if (path == this.TEMPLATE_PATH) {
+      content = this._processTemplateFile(filename, path, content);
+    }
+    this._createFormInput(`files[${filename}]`, this._appendCopyright(filename, content));
+    this._submitForm();
+  }
+
+  _processTemplateFile(filename: string, path: string, content: string): string {
```
jelbourn: How about
```ts
/**
 * The Plunker template assets contain placeholder names for the examples:
 * "<material-docs-example>" and "MaterialDocsExample".
 * This will replace those placeholders with the names from the example metadata, 
 * e.g. "<basic-button-example>" and "BasicButtonExample"
 */
_replaceExamplePlaceholderNames(fileName: string, fileContent: string) { ... }
```
I also noticed that the `path` argument seems to not be used in this function.

tinayuangao:
Comments addressed. Please take another look. Thanks!

src/app/shared/plnkr/index.ts
```diff
@@ -0,0 +1 @@
+export * from './plunker-button';
```
mmalerba: Change the directory name to plunker

src/app/shared/plnkr/plunker-data.ts
```diff
@@ -0,0 +1,16 @@
+/**
+ * Plunker data
+ *   with information about Component name, selector, files used in example, and path to examples
+ */
+export class PlunkerData {
```
mmalerba: I think this is fine for this CL, but this stuff isn't actually related to plunker at all, it should be called ExampleData and should probably live in the material2 repo eventually.

src/app/shared/plnkr/template/index.html
mmalerba: I still advocate moving the template files to assets, they're not really part of the application source

src/app/shared/plnkr/plunker-data.ts
tinayuangao: Changed the name to ExampleData. I think we can extract information (selector, componentName) from example file. I will work on that in another PR.

tinayuangao:
Added test for plunker-writer

jelbourn:
LGTM

jelbourn:
Actually, I just noticed that `NavBar` is now declared in two modules; it should be removed from `AppModule`

tinayuangao:
Updated - fixed NavBar and integrated with example view. Moved example-data to a  better place

jelbourn:
LGTM, we can update to use the real examples in a follow-up PR.

PR #28, opened 2016-11-22
Initial pass of theme chooser

mmalerba:


mmalerba:
@jelbourn I'm having trouble getting the theme css files to actually load. I think the problem is that they're not in my dist directory, but I don't know how to set up the angular-cli to copy them there. Adding it to assets in angular-cli.json doesn't seem to work.

src/app/shared/navbar/navbar.scss
```diff
@@ -9,3 +11,7 @@
   height: 26px;
   vertical-align: middle;
 }
+
+.flex-spacer {
```
jelbourn: Can you prefix this with `docs-`?

(here and elsewhere)

src/app/shared/style-manager/style-manager.ts
```diff
@@ -0,0 +1,26 @@
+import {Injectable} from '@angular/core';
+
+
+@Injectable()
+export class StyleManager {
```
jelbourn: Add class and method descriptions

src/app/shared/theme-chooser/theme-chooser.html
```diff
@@ -0,0 +1,16 @@
+<button md-icon-button [md-menu-trigger-for]="themeMenu">
+  <md-icon>format_color_fill</md-icon>
+</button>
+
+<md-menu class="theme-chooser-menu" #themeMenu="mdMenu" x-position="before">
```
jelbourn: I pulled this down to try it out and found that the keyboard shortcuts for the menu don't work.

src/app/shared/navbar/navbar.scss
mmalerba: does it make sense to prefix the classes if we're not prefixing the components? shouldn't we prefix everything or nothing

jelbourn: It's mainly so that people who inspect the source of the docs site see which styles come from the docs and which are from material itself.

src/app/shared/theme-chooser/theme-chooser.html
jelbourn: Tab should close the menu and the arrow keys should move between items. The menu only becomes focused if you opened it via keyboard, though. If I open the theme picker via keyboard, seems like focus never moves into the menu. Possible that it's a bug in `md-menu`. 

cc @kara

PR #31, opened 2016-12-09
Fix karma test running with `ng test`

jelbourn:
I updated the version of angular-cli and copied `karma.conf.js` and `test.ts` from a fresh `ng new`.


DevVersion:
It looks like you forgot to change the paths in `angular-cli.json` file.

DevVersion:
Everything except `e2e` passes now. Protractor is unable to locate the spec files due to an incorrect path. (See [here](https://github.com/jelbourn/material.angular.io/blob/39ba8085857026b105b2bad03f0419935999960c/protractor.conf.js#L10))

I think once that's fixed, everything should work.

jelbourn:
@DevVersion thanks, you found that before I even checked back to see if it worked.

PR #33, opened 2016-12-19
Expand component view with example viewer

jelbourn:
* Add `DocumentationItems` service
* Add content to `ComponentView` sidenav
* Rebase and add functionality to `ExampleViewer`
* Add script to syntax-highlight example assets

I want to get this PR in so that @amitafr can start working on styles. Currently working on unit tests for `ExampleViewer`


mmalerba:
lgtm, I have an updated version of the button (and other) examples, but we can always update those in a follow up pr

src/app/examples/example-module.ts
```diff
@@ -0,0 +1,34 @@
...
+export const EXAMPLE_COMPONENTS = {
+  'button-overview': ButtonOverviewExample,
+  'button-types': ButtonTypesExample
+};
+
+/**
+ * The list of all example components.
+ * We need to put them in both `declarations` and `entryComponents` to make them work.
+ */
+export const EXAMPLE_LIST = [
```
tinayuangao: get the list from EXAMPLE_COMPONENT so we don't have to write them twice?

jelbourn: We can't- doing that would be imcompatiblr with AoT because the values have to be static constants.

PR #49, opened 2016-12-22
Add missing examples, example titles, and ensure examples are destroyed

jelbourn:
* Add missing examples
* Add example titles
* Fix examples not being unloaded when doc-viewer is destroyed
* Update overview static content


crisbeto:
LGTM, don't have permission to re-label and review stuff here.

PR #52, opened 2016-12-22
fix: example-viewer icons and animation

EladBezalel:
- changed codepen icon to a generic icon
- added open animation
- added tooltips


jelbourn:
LGTM

PR #53, opened 2016-12-22
fix: add plainText option to doc-viewer so html in example-viewer won't be rendered

EladBezalel:


jelbourn:
I already wrote a script for doing this with hljs, closing since that's now used.

PR #55, opened 2016-12-22
Clean up examples and fix a few loading bugs

jelbourn:
* Add css files for all examples
* Fix dialog and snackbar examples not to cause an action on page-load
* Fix example-viewer sometimes dropping to the bottom of the page
* Fix source path for example assets
* Generate syntax-highlighted example sources
* Update API docs and overviews


tinayuangao:
LGTM

PR #56, opened 2016-12-22
Fix plunker: rename radio-ngmodel; add FormsModule; use css

tinayuangao:
- Renamed radio-ngmodel to radio-ng-model so the component name matches the selector
- Add `FormsModule` so `ngModel` is working now
- Change the example css file from `.scss` to `.css` 

R: @jelbourn


jelbourn:
LGTM

PR #57, opened 2016-12-22
fix: app background cut off

crisbeto:
* Fixes the app background being cut off below the fold.
* Removes the "Loading..." sign while loading.


jelbourn:
LGTM

PR #60, opened 2016-12-22
Fix asset urls for examples that load images/icons

jelbourn:


src/app/examples/card-fancy/card-fancy-example.css
```diff
@@ -3,6 +3,6 @@
 }
 
 .example-header-image {
-  background-image: url('shiba1.jpg');
+  background-image: url('assets/img/examples/shiba2.jpg');
```
tinayuangao: Should it be shiba1?

jelbourn: Changed

PR #63, opened 2016-12-22
Correct some typos.

josephperrott:


jelbourn:
Fixed in the source of truth.

PR #70, opened 2016-12-23
Fix plunkers and add tabs to the site

jelbourn:
* Fixes the plunker template by updating the umd bundle path and copying the example sources under `assets/` (only the `assets/` folder sticks around for static servers like firebase).
* Add tabs, which were missing
* Remove implementation from tabs extra example because it was throwing an animation error (despite still working) 
* Bump the angular version by one patch
* Regenerate some static assets


mmalerba:
lgtm

PR #75, opened 2017-01-04
fix(main.scss): font-weight property

zalog:
This was just a little glitch.


jelbourn:
Fixed in a previous PR

PR #76, opened 2017-01-04
Publish scripts

ericjim:
It is assumed that the deployer has access to both instances (dev and prod)

### DIFF
--- a/.firebaserc
+++ b/.firebaserc
@@ -1,6 +1,1 @@
 {}
-{
-  "projects": {
-    "default": "material2-docs-dev"
-  }
-}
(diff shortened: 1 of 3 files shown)


.firebaserc
```diff
@@ -1,6 +1 @@
 {}
-{
```
ericjim: The command `firebase deploy` defaults to `material2-docs-dev`, however, we can control this by adding the instance name as a parameter (refer to `publish-dev`).

jelbourn:
LGTM

We'll eventually want to have this command also pull down the appropriate static assets

PR #80, opened 2017-01-10
Guides list

amitafr:
Add Guides into Table List layout


jelbourn:
Folding this into #144

PR #81, opened 2017-01-12
API reference text tweaks

amitafr:
- Colour @input @output markers
- Style Method headers 
- Adjust Parameters and Returns text


jelbourn:
Absorbed into #145

PR #82, opened 2017-01-13
fix(scrolling): enable scrolling on safari

### LINKED ISSUE #71: I cannot scroll docs' main content in safari
When I open any doc page, I cannot scroll right zone with content, I can scroll only left zone (with components list).

Env: macOS Sierra 10.12.2, Safari 10.0.2 (12602.3.12.0.1)

UPD. In Firefox left column (with menu list) and right column (with content) scrolls together (must scrolls separately instead).

andrewseguin:
This is a simple one-liner that removes the `md-sidenav-content` from the flow of `md-sidenav-layout` so that layout does not derive its scroll height from the content. 

On Safari with this fix, the `md-sidenav-content` matches its bounding box client rect height and scroll height to the layout which now is correctly flex'ed to the rest of the page's height. 

Closes #71

### DIFF
--- a/src/app/pages/component-sidenav/component-sidenav.scss
+++ b/src/app/pages/component-sidenav/component-sidenav.scss
@@ -7,6 +7,10 @@ app-component-sidenav {
 
 .docs-component-viewer-sidenav-container {
   width: 100%;
+  
+  .md-sidenav-content {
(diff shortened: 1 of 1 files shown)


jelbourn:
LGTM

PR #83, opened 2017-01-13
fix(plunker): make calls to plunker synchronous to avoid popup block

andrewseguin:


src/app/shared/plunker/plunker-button.ts
```diff
@@ -6,18 +6,30 @@ import { ExampleData } from '../../examples/example-data';
...
+    '(mouseover)': 'isDisabled = !plunkerForm'
+  }
 })
 export class PlunkerButton {
-  exampleData: ExampleData;
+  /**
+   * The button becomes disabled if the user hovers over the button before the plunker form
+   * is created. After the form is created, the button becomes enabled again.
+   */
+  isDisabled: boolean = false;
```
tinayuangao: Should we make it disabled by default?

andrewseguin: Initially I had it disabled by default and only enabled when the form was created. However, this caused a bit of a flicker on the button where it goes from disabled -> enabled right after the page loads.

With the current implementation, it always appears active to the user except in the small chance they manage to mouse over it to interact with it. In which case, it disables itself and shows a tooltip saying that we building the form.

tinayuangao: LGTM

andrewseguin:
Added a comment about the speed of creating the form

Issue #88, opened 2017-01-26
'md-sidenav-layout' used in page 'component-sidenav' and cause page load error


weichen2046:
According to the `sidenav` [README](https://github.com/angular/material2/blob/master/src/lib/sidenav/README.md) should use `md-sidenav-container` instead.

>> NOTE: The md-sidenav-layout element is deprecated. md-sidenav-container should be used instead.

[README]:https://github.com/angular/material2/blob/master/src/lib/sidenav/README.md

andrewseguin:
Fixed in #102, thanks

PR #89, opened 2017-01-26
fix(sidenav): use md-sidenav-container instead of md-sidenav-layout (…

weichen2046:
…#88)


weichen2046:
I signed it!

andrewseguin:
Fixed in #102, thanks

PR #91, opened 2017-01-30
fix(readability): changed text opacity from .54 to .87 according the spec

EladBezalel:
fixes https://github.com/angular/material2/issues/2689

### DIFF
--- a/src/app/pages/component-viewer/_component-viewer-theme.scss
+++ b/src/app/pages/component-viewer/_component-viewer-theme.scss
@@ -8,10 +8,10 @@
   $foreground: map-get($theme, foreground);
 
   app-component-viewer {
-    color: mat-color($foreground, secondary-text);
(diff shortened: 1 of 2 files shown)


EladBezalel:
| Before | After |
| ------- | ----- |
| image | image  |

jelbourn:
I think the white-on-blue of the header also needs to be changed, as it's pretty low ratio now as well.

EladBezalel:
@jelbourn @kara @amitafr - regarding the header - it seems like the `font-weight` right now is `300`, should i change it to `400` or do you have anything else in mind?

andrewseguin:
Hey Elad, can you rebase this change

EladBezalel:
@andrewseguin @kara rebased :)

davidglezz:
Text still at .54

Issue #111, opened 2017-02-24
A11y Screen Reader - Page structure challenges


DaveBest99:
Screen reader WCAG issues.
1. Perceivable: Page structure lacks content relationships; (no <main> region), (missing text label on <nav>), (missing text label on <footer>).
2. Operable: Pressing Enter key on Link elements (like "COMPONENTS" and "GUIDES") provide no verbal feedback of cursor focus change to notify the blind user.
3. Understandable: The page is simple but is missing Breadcrumb and proper page Title for orientation.
4. Robust: Not tested.
5. See attached files.

[Angular Accessibility Introduction.docx](https://github.com/angular/material.angular.io/files/800324/Angular.Accessibility.Introduction.docx)

Angular Material Accessibility Introduction.mp4: https://www.dropbox.com/s/rz2gtuhnbx9czus/Angular%20Material%20Accessibility%20Introduction.mp4?dl=0

naomiblack:
@riavalon could you update this issue with status? Other than the issue with routing, where are we at on the remainder of these page structure questions?

riavalon:
Sure thing: I've added in some tweaks to update how the page is structured with different regions as well as added the ability to have the page title updated on route change now. There is still an issue with getting verbal feedback about page transitions that I don't think we have officially settled on a solution for. There's more info in the PR I created: #117

PR #117, opened 2017-03-02
chore(accessibility): Add aria attributes to aid screen reader flow

riavalon:
This is to address #111 issues


src/app/pages/component-category-list/component-category-list.html
```diff
@@ -1,5 +1,6 @@
 <div class="docs-component-category-list">
   <md-card
+      role="link"
```
jelbourn: I think a better approach for this one would be to wrap the entire card inside of an anchor element.

src/app/pages/component-sidenav/component-sidenav.html
```diff
@@ -1,9 +1,9 @@
 <md-sidenav-container class="docs-component-viewer-sidenav-container">
-  <md-sidenav #sidenav class="docs-component-viewer-sidenav"
+  <md-sidenav aria-label="Side navigation" role="navigation" #sidenav class="docs-component-viewer-sidenav"
```
jelbourn: I'm wondering if we should avoid having `<nav>` inside of a `role="navigation"` and instead wrap the entire sidenav content in one `<nav>`

jelbourn: A more informative aria-label might be "Component navigation"

src/app/pages/component-viewer/component-viewer.html
```diff
@@ -1,22 +1,24 @@
 <!-- TODO(jelbourn): turn this into nav tabs -->
-<md-tab-group class="docs-component-viewer-tabbed-content"
+<md-tab-group role="tablist" class="docs-component-viewer-tabbed-content"
```
jelbourn: The roles and other aria attributes are something that really belong as part of the md-tab-group implementation

src/app/pages/guide-list/guide-list.scss
```diff
@@ -1,4 +1,4 @@
-:host {
+:host, .guides-container {
```
jelbourn: Should be prefixed as `.docs-guide-container`

src/app/shared/documentation-items/documentation-items.ts
```diff
@@ -29,7 +29,7 @@ const DOCS = [
   },
   {
     id: 'nav',
-    name: 'Navigation',
+    name: 'Navigation Components',
```
jelbourn: It's somewhat redundant for this to say "Navigation components" since they're all components. Could it perhaps just be done as an aria label?

src/app/shared/example-viewer/example-viewer.html
```diff
@@ -3,6 +3,8 @@
     <div class="docs-example-viewer-title-spacer">{{exampleData?.title}}</div>
 
     <button md-icon-button type="button" (click)="toggleSourceView()"
+            role="button"
```
jelbourn: `role="button"` should never be necessary on a `<button>` element

(here and elsewhere)

src/app/shared/focus-element/focus-element.ts
```diff
@@ -0,0 +1,26 @@
...
+  ngAfterViewInit() {
+    // Add tabindex 0 so element is focusable by keyboard
+    this
+      ._renderer
+      .setElementAttribute(this._element.nativeElement, 'tabindex', '0');
+
+    // Focus host element after view loads
+    // so screen readers will be alerted
+    // the page has changed.
+    this._element.nativeElement.focus();
```
jelbourn: This seems like a very risky change to make; generally having an element grab focus to itself without user input can cause problems. I'll try to look into alternate approaches

src/app/shared/navbar/navbar.html
```diff
@@ -1,12 +1,12 @@
 <!-- TODO: figure out if the <nav> should go inside of a <header> element. -->
-<nav class="docs-navbar">
+<nav aria-label="Main" role="navigation" class="docs-navbar">
```
jelbourn: I'm not sure `role="navigation"` should be necessary for a `<nav>` element

src/app/shared/navbar/navbar.html
```diff
@@ -1,12 +1,12 @@
...
   <a md-button routerLink="/" aria-label="Angular Material">
     <img class="docs-angular-logo"
          src="../../../assets/img/homepage/angular-white-transparent.svg"
          alt="Angular">
     <span>Material</span>
   </a>
-  <a md-button routerLink="components">Components</a>
-  <a md-button routerLink="guides">Guides</a>
+  <a aria-label="Components" role="link" md-button routerLink="components">Components</a>
+  <a aria-label="Guides" role="link" md-button routerLink="guides">Guides</a>
```
jelbourn: `role="link"` shouldn't be necessary for an `<a>` element

src/app/pages/component-category-list/component-category-list.html
riavalon: We would have to add an `<a role="link">` around the contents of each card, though. Isn't it a bit more concise to just add the role-link to the md-card itself?

src/app/shared/documentation-items/documentation-items.ts
riavalon: Ah, I just added "Components" to it since the screen reader otherwise reads it as "Navigation navigation". I'll see if we can change it to an aria label instead. 👍

src/app/pages/component-category-list/component-category-list.html
jelbourn: In general it's always better to use the appropriate native element instead of a role. An `<a>` also has the proper associated `href` and tab index.

src/app/pages/component-sidenav/component-sidenav.html
riavalon: I don't think that this is necessarily a problem. To my knowledge, a `<nav>` element or an element with `role="navigation"` just cannot have a `<main>` or `role="main"` as a descendant. Adding the role navigation here to allow screen readers to see the whole side nav, and then dig down into the individual sections of the side nav was suggested based off this issue: https://github.com/angular/material.angular.io/issues/111

src/app/shared/focus-element/focus-element.ts
naomiblack: @riavalon have you tried setting a document title? I would expect a screen reader to notice and announce a change in the title. Have a look here: https://angular.io/docs/ts/latest/cookbook/set-document-title.html
Can you see if this is a better solution?

riavalon: I actually tried this already, even with a title change the screen reader does not announce anything else.

naomiblack: that is definitely weird. I'll see if I can get some help on this one -- yes -- please do add a document title change back in.

DaveBest99:
Issue:
Page content should not update without being initiated by a user action, or some notification is given to the user (2.4 Alert and Message Dialogs) to inform of the page behaviour.
On the Angular Material demo page (https://material.angular.io/components) the page content is updated when the user presses the Enter key on one of the Navigation Link elements. However, the screen reader does not notify the user that page content has been updated.

User action:
- Press the Tab key to move focus to a Navigation Link element "Autocomplete".
- Press Enter key to activate the Link element.
- Page content updates, but screen reader focus does not move and no screen reader output announces the page change.

Solutions:
1. Do not move the cursor focus, but announce the page update by adding an aria-alert to the h1 tag "Component - Autocomplete" (2.3 ARIA Alert role).
2. After the page content updates, upon the Enter key press, move the cursor focus to the h1 tag "Component - Autocomplete" (2.14 ARIA Link - Enter: Executes the link and moves focus to the link target).
3. Update the page Title by appending the h1 tag "Component - Autocomplete" to the Title (I.E. "Component - Autocomplete - Angular Material").

Notes:

- WCAG2.4.2 Page Titled: Web pages have titles that describe topic or purpose. (Level A).
The intent of this Success Criterion is to help users find content and orient themselves within it by ensuring that each Web page has a descriptive title. Titles identify the current location without requiring users to read or interpret page content.
http://www.w3.org/TR/UNDERSTANDING-WCAG20/complete.html

- ARIA2.5 Breadcrumb: A breadcrumb trail consists of a list of links to the parent pages of the current page in hierarchical order. It helps users find their place within a website or web application. Breadcrumbs are often placed horizontally before a page's main content. 
http://w3c.github.io/aria-practices/

PR #129, opened 2017-03-08
Update asset fetch script to use real repo

jelbourn:


tools/fetch-assets.sh
```diff
@@ -3,31 +3,36 @@
...
+
+# Create folders into which to copy content and assets.
+mkdir -p $tmpAssetClonePath
+mkdir -p $exampleAssetsPath $docAssetsPath
 
 # Pull assets from repo and remove .git directory
-git clone $DOCS_REPO $DEST_PATH
-rm -rf $DEST_PATH/.git
+git clone $docsContentRepo $tmpAssetClonePath
+rm -rf $tmpAssetClonePath/.git
```
DevVersion: Why do we need to remove the `.git/` folder?

jelbourn: No idea- @riavalon what was the reason for that?

riavalon: Honestly, I'm not quite sure anymore. It doesn't seem like it's needed, though, so it should be removed.

DevVersion: We could remove it I think. Other than that this PR should be ready?

PR #134, opened 2017-03-17
feat(navbar): Add basic search bar component to site.

riavalon:


src/app/shared/navbar/searchbar/searchbar.ts
```diff
@@ -0,0 +1,51 @@
...
+    private _snackBar: MdSnackBar
+  ) {
+    this.allDocItems = _docItems.getAllItems();
+    this.filteredSuggestions = this.searchControl.valueChanges
+      .startWith(null)
+      .map(item => this.filterSearchSuggestions(item));
+  }
+
+  public toggleIsExpanded() {
+    this._isExpanded = !this._isExpanded;
```
kara: The autocomplete trigger has a `panelOpen` property. Have you tried using that instead of toggling it here? You should be able to do a ViewChild query for `'auto'` to get a reference to the trigger.

```
host: {
   '[class.expanded]': '_autocompleteTrigger?.panelOpen'
}
```

src/app/shared/navbar/searchbar/searchbar.ts
```diff
@@ -0,0 +1,51 @@
...
+
+
+@Component({
+  selector: 'search-bar-component',
+  templateUrl: './searchbar.html',
+  styleUrls: ['./searchbar.scss']
+})
+
+export class SearchBar {
+  @HostBinding('class.expanded') _isExpanded = false;
```
kara: Move this to the host object in the Component metadata? Generally prefer that instead of using the decorator.

src/app/shared/navbar/searchbar/searchbar.html
```diff
@@ -0,0 +1,16 @@
+<div class="search-input-container">
+  <input
+    placeholder="Search"
+    type="text"
+    (focus)="toggleIsExpanded(true)"
+    (blur)="toggleIsExpanded(false)"
+    (keyup.enter)="navigate($event.target.value.toLowerCase())"
```
kara: Does this also navigate when you click on the option?Also have you considered switching to the `optionSelections` observable on the trigger rather than this event? It emits any option object that is selected (via the keyboard or the mouse).

src/app/shared/navbar/searchbar/searchbar.html
```diff
@@ -0,0 +1,16 @@
...
+    [mdAutocomplete]="auto"
+    [formControl]="searchControl">
+</div>
+
+<md-autocomplete #auto="mdAutocomplete">
+  <md-option *ngFor="let item of filteredSuggestions | async" [value]="item.name">
```
kara: Another idea would be to bind the entire item as the option value, and use the `displayWith` property on `md-autocomplete` to map that object to its display value (the name). That way, you don't have to go through the list in `navigate()` to find the ID that matches the search string. You can just get the object itself and pass through its ID.

src/app/shared/navbar/searchbar/searchbar.ts
```diff
@@ -0,0 +1,51 @@
+import {Component, HostBinding} from '@angular/core';
+import {Router} from '@angular/router';
+import {FormControl} from '@angular/forms';
+import {MdSnackBar} from '@angular/material';
+import {Observable} from 'rxjs';
```
kara: Can you restrict the import a bit? From `'rxjs/Observable'`

src/app/shared/navbar/searchbar/searchbar.scss
```diff
@@ -0,0 +1,62 @@
...
+
+:host {
+  position: relative;
+  flex: 2;
+
+  * {
+    box-sizing: border-box;
+  }
+
+  &.expanded .search-input-container {
```
jelbourn: All of the css should start with a `docs-` prefix
(mainly so that people looking at the docs site in dev tools can differentiate between docs-site-specific css and the underlying library css)

src/app/shared/navbar/searchbar/searchbar.scss
```diff
@@ -0,0 +1,62 @@
...
+    padding-left: 50px;
+    position: relative;
+    transition: width .2s ease;
+    width: 100%;
+
+    /* Set placeholder text to be white */
+    &::-webkit-input-placeholder { @include color-placeholder(); } /* Chrome/Opera/Safari */
+    &::-moz-placeholder { @include color-placeholder(); } /* Firefox 19+ */
+    &:-moz-placeholder { @include color-placeholder(); } /* Firefox 18- */
+    &:ms-input-placeholder { @include color-placeholder(); } /* IE 10+ */
```
jelbourn: autoprefixer is supposed to be built into the angular-cli, but I'm not sure if there's some specific configuration needed to turn it on

src/app/shared/navbar/searchbar/searchbar.ts
jelbourn: To elaborate: when you import from just `rxjs`, it brings in a lot of than just what you're using and increases the payload size.

src/assets/img/icons/ic_search_white_24px.svg
```diff
@@ -0,0 +1,4 @@
+<svg fill="#FFFFFF" height="24" viewBox="0 0 24 24" width="24" xmlns="http://www.w3.org/2000/svg">
+    <path d="M15.5 14h-.79l-.28-.27C15.41 12.59 16 11.11 16 9.5 16 5.91 13.09 3 9.5 3S3 5.91 3 9.5 5.91 16 9.5 16c1.61 0 3.09-.59 4.23-1.57l.27.28v.79l5 4.99L20.49 19l-4.99-5zm-6 0C7.01 14 5 11.99 5 9.5S7.01 5 9.5 5 14 7.01 14 9.5 11.99 14 9.5 14z"/>
+    <path d="M0 0h24v24H0z" fill="none"/>
+</svg>
```
jelbourn: Is there a font icon for search that would serve for this instead of using an svg?

src/app/shared/navbar/searchbar/searchbar.scss
riavalon: I did see that it was auto-prefixing some styles, but the input placeholder one unfortunately wasn't being done for some reason.

src/app/shared/navbar/searchbar/searchbar.ts
riavalon: I added it as a decorator as the docs on angular.io suggest it's better to prefer the decorator over the metadata: https://angular.io/docs/ts/latest/guide/style-guide.html#!#06-03

kara: In Material, we actually do the opposite because @HostBinding causes issues with Angular Universal (which it seems the docs authors weren't aware of)

riavalon: Ahh, I see. Alrighty, I'll switch it up 👍

src/app/shared/navbar/searchbar/searchbar.ts
riavalon: I believe this will cause the input to expand only when the user starts typing and suggesting show up. We want the input to enlarge right when the user focuses it so I'm using the `_isExpanded` here instead

kara: The autocomplete panel should appear whenever the input is focused. If that's not happening, seems like an autocomplete bug.

riavalon: I think I may be. It seems that the first time the field is focused it does not show the panel, and the second time the field is focused it does show it.

riavalon:
@jelbourn This should be ready for another look over again

src/app/shared/navbar/searchbar/searchbar.html
```diff
@@ -0,0 +1,17 @@
+<div class="docs-search-input-container">
+  <input
+    placeholder="Search"
+    type="text"
+    (focus)="toggleIsExpanded(true)"
+    (blur)="toggleIsExpanded(false)"
```
jelbourn: This method doesn't take an argument

src/app/shared/navbar/searchbar/searchbar.scss
```diff
@@ -0,0 +1,60 @@
+$input-bg: #85D9E2;
```
jelbourn: All of the colors should be part of the theme

src/app/shared/navbar/searchbar/searchbar.html
```diff
@@ -0,0 +1,17 @@
...
+    (keyup.enter)="handlePlainSearch($event.target.value.toLowerCase())"
+    [mdAutocomplete]="auto"
+    [formControl]="searchControl">
+  <i class="material-icons">search</i>
```
jelbourn: This should probably be `<md-icon>`

src/app/shared/navbar/searchbar/searchbar.scss
```diff
@@ -0,0 +1,60 @@
...
+    i.material-icons {
+      position: absolute;
+      left: 15px; top: 50%;
+      transform: translateY(-50%);
+      height: 28px;
+      width: 28px;
+    }
+  }
+
+  input {
```
jelbourn: Can this be done with a css class instead? E.g. `.docs-search-input`

src/app/shared/navbar/searchbar/searchbar.ts
```diff
@@ -0,0 +1,80 @@
...
+  public displayFn(item: DocItem) {
+    return item.name;
+  }
+
+  public filterSearchSuggestions(searchTerm): DocItem[] {
+    return this.allDocItems.filter(item => new RegExp(`^${searchTerm}`, 'gi').test(item.name));
+  }
+
+  public handlePlainSearch(searchTerm) {
+    const item = this.allDocItems.find(item => item.name.toLowerCase() === searchTerm);
```
jelbourn: Partial matches should also just to the first result. E.g., typing "Date" and hitting enter should jump to the datepicker.

src/app/shared/navbar/searchbar/searchbar.ts
```diff
@@ -0,0 +1,80 @@
...
+    item ? this.navigate(item.id) : this._showError();
```
jelbourn: We should clear the input on a successful search

src/app/shared/navbar/searchbar/searchbar.ts
```diff
@@ -0,0 +1,80 @@
...
+    '[class.docs-expanded]': '_isExpanded'
+  }
+})
+
+export class SearchBar {
+
+  @ViewChild(MdAutocompleteTrigger)
+  private _autocompleteTrigger: MdAutocompleteTrigger;
+
+  public allDocItems: DocItem[];
```
jelbourn: Don't need to explicitly say `public` since that's the default

src/app/shared/navbar/searchbar/searchbar.ts
```diff
@@ -0,0 +1,80 @@
...
+  public filteredSuggestions: Observable<DocItem[]>;
+  public searchControl: FormControl = new FormControl('');
+  public sub;
```
jelbourn: `sub` should have a more specific name and probably a type

riavalon:
@jelbourn Updated this with the recent requested changes

jelbourn:
Looks good except for one last thing I noticed:
image

When the input is first focused, the autocomplete uses the original width before the input finishes expanding. In reality this is probably a while new feature we need in autocomplete, but for now I think we can work around it by immediately closing the autocomplete on focus and then opening it manually when the input's expansion animation completes.

riavalon:
@jelbourn Set the panel to open and close in sync with the expanding animation

jelbourn:
@riavalon looks like clicking on the autocomplete items doesn't navigate any more

riavalon:
@jelbourn Blur event was being called when trying to click on an option. I rewrote the focus/blur logic to ignore clicks on `md-option`s.

jotatoledo:
Up, this would be a nice feature on the docs site

PR #138, opened 2017-03-30
chore(tests): Add unit tests for each component.

riavalon:


jelbourn:
Looks like CI is failing on an scss issue

riavalon:
I believe the scss issue was due to it building with `@angular/material` set to beta 1 instead of 2. There is an issue with one of the tests now, and I'm looking into it 👍

package.json
```diff
@@ -21,7 +21,7 @@
     "@angular/core": "^2.4.6",
     "@angular/forms": "^2.4.6",
     "@angular/http": "^2.4.6",
-    "@angular/material": "^2.0.0-beta.1",
+    "@angular/material": "^2.0.0-beta.2",
```
jelbourn: Can revert this line now that we're on beta.3

src/app/pages/homepage/homepage.spec.ts
```diff
@@ -0,0 +1,40 @@
...
+      imports: [MaterialModule, RouterTestingModule],
+      schemas: [CUSTOM_ELEMENTS_SCHEMA],
+      declarations: [Homepage]
+    });
+
+    fixture = TestBed.createComponent(Homepage);
+    fixture.detectChanges();
+  }));
+
+  it('should render four rows of content on the homepage', () => {
```
jelbourn: I don't think we need these homepage tests since it's all just static content.

src/app/shared/footer/footer.spec.ts
```diff
@@ -0,0 +1,40 @@
...
+    fixture.detectChanges();
+  }));
+
+  it('should have a link to angular.io', () => {
+    const link = fixture
+      .nativeElement
+      .querySelector('.docs-footer-links a');
+    const href = link.getAttribute('href');
+    const text = link.textContent;
+    expect(~href.indexOf('angular.io')).toBeTruthy();
```
jelbourn: Should be able to do
```ts
expect(href).toContain('angular.io');
```

src/app/shared/footer/footer.spec.ts
```diff
@@ -0,0 +1,40 @@
...
+    expect(text).toContain('Learn Angular');
+  });
+
+  it('should show an angular logo in the footer', () => {
```
jelbourn: I don't think a test for the this is necessary

src/app/shared/navbar/navbar.spec.ts
```diff
@@ -0,0 +1,37 @@
...
+      schemas: [CUSTOM_ELEMENTS_SCHEMA],
+      declarations: [NavBar],
+    });
+
+    fixture = TestBed.createComponent(NavBar);
+    component = fixture.componentInstance;
+    fixture.detectChanges();
+  }));
+
+  it('should have four main links', () => {
```
jelbourn: I think this test also isn't needed

PR #142, opened 2017-04-06
Bug (Snack-bar / snackbar): fixed snackbar image not showing up

M-a-c:
The snackbar svg is not 'snack-bar', but 'snackbar'.

Additional question.
* Snack-bar on Material.io is always one word never with a hyphen in it. a lot of times when it is referenced in this and the Angular/material2 repo its referenced as 'snack-bar' quite a bit. Is it correct either way, or should they all be either (snackbar/snack-bar) ?

snackbar


josephperrott:
Thanks for providing this, a fix was merged for this in #183

PR #144, opened 2017-04-08
Use md-nav-list for guide list

jelbourn:
Adds on top of #80, but I remove the `<table>` since it's not quite right to use that inside of a list. 

I also changed the `h1` style to be regular text.


src/app/pages/guide-list/_guide-list-theme.scss
```diff
@@ -0,0 +1,13 @@
+@import '~@angular/material/theming';
+
+@mixin guide-list-theme($theme) {
+  $primary: map-get($theme, primary);
```
crisbeto: Most of these variables aren't being used, perhaps we can get rid of them?

PR #162, opened 2017-05-31
bug(progress-spinner): color property binding fix

girtsernstsons:
Radio button for color/theme switch isn't working as color property was specified as [attr.color] instead of [color]


girtsernstsons:
I signed it!

girtsernstsons:
Closed as is fixed in https://github.com/angular/material2/pull/4687

PR #163, opened 2017-06-01
Change fetch-assets script to work w/ npm 5

jelbourn:


jelbourn:
Yeah, not a fan of that change

PR #203, opened 2017-06-30
fix: site crashing on browsers with limited localStorage access

crisbeto:
Prevents the site from crashing on browsers that don't allow access to localStorage (e.g. IE11 or iOS Safari in private mode).

### DIFF
--- a/src/app/shared/theme-picker/theme-storage/theme-storage.ts
+++ b/src/app/shared/theme-picker/theme-storage/theme-storage.ts
@@ -6,7 +6,7 @@ export interface DocsSiteTheme {
   primary: string;
   isDark?: boolean;
   isDefault?: boolean;
-};
+}
 
 
 @Injectable()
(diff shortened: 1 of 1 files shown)


src/app/shared/theme-picker/theme-storage/theme-storage.ts
```diff
@@ -16,15 +16,24 @@ export class ThemeStorage {
   public onThemeUpdate: EventEmitter<DocsSiteTheme> = new EventEmitter<DocsSiteTheme>();
 
   public storeTheme(theme: DocsSiteTheme) {
-    window.localStorage[ThemeStorage.storageKey] = JSON.stringify(theme);
+    try {
+      window.localStorage[ThemeStorage.storageKey] = JSON.stringify(theme);
```
jelbourn: Can you file an issue to fall-back to session storage for these cases?

jelbourn:
LGTM

src/app/shared/theme-picker/theme-storage/theme-storage.ts
crisbeto: I don't think `sessionStorage` would be available in this case either (at least it doesn't look like it is on IE11).

jelbourn: Ack, too bad.

PR #205, opened 2017-07-01
chore(style): added polish to code snippet examples

amcdnl:
Polishes the inline code example styles by adding ascent colors and padding.

**light**
light

**dark**
dark

### DIFF
--- a/src/styles/_markdown-theme.scss
+++ b/src/styles/_markdown-theme.scss
@@ -12,5 +12,10 @@
   .docs-markdown-a {
     color: mat-color($primary);
   }
+
+  .docs-markdown-pre {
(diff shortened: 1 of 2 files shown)


amcdnl:
@jelbourn - I think in another PR i'd like to convert the highlighter theme to actual material. Thoughts?

jelbourn:
@amcdnl yeah, that would be great. We just grabbed off-the-shelf themes for expedience

jelbourn:
(it's there's an existing material theme that would be even better)

amcdnl:
@jelbourn - There is a dark material one ( https://github.com/Kelbster/highlightjs-material-dark-theme ) but not a light one.

PR #209, opened 2017-07-05
fix: include proper font and web animations polyfill

crisbeto:
* Includes the `Roboto` font from Google Fonts. Previously only `Roboto Mono` was included.
* Includes the Web animations polyfill in order for the component animations to work on all browsers.

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -16,7 +16,7 @@
       sizes="192x192">
   <link rel="manifest" href="assets/img/favicons/manifest.json">
   <link href="https://fonts.googleapis.com/icon?family=Material+Icons" rel="stylesheet">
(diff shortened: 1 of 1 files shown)


src/index.html
```diff
@@ -28,5 +28,6 @@
   </script>
   <script async src='https://www.google-analytics.com/analytics.js'></script>
   <script src="https://ajax.googleapis.com/ajax/libs/hammerjs/2.0.8/hammer.min.js"></script>
+  <script src="https://cdnjs.cloudflare.com/ajax/libs/web-animations/2.2.5/web-animations.min.js"></script>
```
jelbourn: Would it make sense to make this script `async`, think that it would be loaded by the time someone interacts with something animated?

crisbeto: I'm not sure how that would work out if the request takes too long and Angular is bootstrapped by then.

PR #213, opened 2017-07-08
chore(styles): apply material light / dark styles to highlightjs

amcdnl:
Update highlightjs to use material style syntax for light and dark themes.

**light**
light

**dark**
dark

### DIFF
--- a/src/highlightjs/material-dark.css
+++ b/src/highlightjs/material-dark.css
@@ -1,11 +1,13 @@
-/** From https://github.com/Kelbster/highlightjs-material-dark-theme */
+/*
+Orginal Style from https://github.com/Kelbster/highlightjs-material-dark-theme  (c) Kelby Gassmanl <kelby.gassman@gmail.com>
+*/
 .hljs {
   display: block;
   overflow-x: auto;
   padding: 1em;
   background: #2B2B2D;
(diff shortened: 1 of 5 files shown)


src/highlightjs/material-dark.css
```diff
@@ -0,0 +1,114 @@
+/** From https://github.com/Kelbster/highlightjs-material-dark-theme */
```
jelbourn: Can we consume this from npm? Just copying like this would require the license to be included as well.

src/highlightjs/material-dark.css
```diff
@@ -0,0 +1,114 @@
...
+.hljs {
+  display: block;
+  overflow-x: auto;
+  padding: 1em;
+  background: #2B2B2D;
+  color: #CDD3D8;
+  font-family: "Fira Mono", monospace;
```
jelbourn: We should use `Roboto Mono` (for both)

src/highlightjs/material-dark.css
amcdnl: I have some slight modifications, I'll reference the license.

amcdnl:
@jelbourn - I've addressed the items and ready for another review.

PR #217, opened 2017-07-11
fix(footer): extra w in www.angular.io

Splaktar:
https://wwww.angular.io -> https://angular.io


rafaelss95:
In fact there's no necessity of the `www` :) can be just https://angular.io/

Splaktar:
@rafaelss95 yep, updated.

Issue #189, opened 2017-06-18
Update page title based on current view


jelbourn:
Current the page title is always "Angular Material"

We should update it to reflect the current view, similar to https://developers.google.com/web/fundamentals/performance. 

This helps people who have multiple tabs open at once.

josephperrott:
Fixed via #220

PR #220, opened 2017-07-12
feat(docs): add page titles to doc pages #189

amcdnl:
Addresses #189 by setting page titles by enhancing the ComponentPageTitle service.

### DIFF
--- a/src/app/pages/page-title/page-title.ts
+++ b/src/app/pages/page-title/page-title.ts
@@ -13,9 +13,9 @@ export class ComponentPageTitle {
   set title(title: string) {
     this._title = title;
     if (title !== '') {
-      title = ` - ${title}`;
+      title = `${title} | `;
     }
-    this.bodyTitle.setTitle(`Angular Material${title}`);
+    this.bodyTitle.setTitle(`${title}Angular Material`);
   }
 
   constructor(private bodyTitle: Title) { }
--- a/src/app/pages/page-title/page-title.ts
+++ b/src/app/pages/page-title/page-title.ts
@@ -1,4 +1,5 @@
 import {Injectable} from '@angular/core';
+import {Title} from '@angular/platform-browser';
 
 /**
  * Service responsible for setting the title that appears above the components and guide pages.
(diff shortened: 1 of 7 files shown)


src/app/pages/page-title/page-title.ts
```diff
@@ -8,5 +9,15 @@ export class ComponentPageTitle {
...
 
   get title(): string { return this._title; }
-  set title(title: string) { this._title = title; }
+
+  set title(title: string) {
+    this._title = title;
+    if (title !== '') {
+      title = ` - ${title}`;
+    }
+    this.bodyTitle.setTitle(`Angular Material${title}`);
```
jelbourn: The order should be the other way around, e.g.
`"Autocomplete | Angular Material"`

This is better because people can have multiple tabs open and can easily see which tab is which without switching to it. Also use `|` instead of `-`
(see this pattern in-action at, e.g., https://developers.google.com/web/fundamentals/performance/prpl-pattern/)

amcdnl:
@jelbourn - ready for further review.

src/app/pages/page-title/page-title.ts
```diff
@@ -13,9 +13,9 @@ export class ComponentPageTitle {
   set title(title: string) {
     this._title = title;
     if (title !== '') {
```
rafaelss95: why not simply `if (title) { ... }`?

src/app/pages/page-title/page-title.ts
```diff
@@ -8,5 +9,15 @@ export class ComponentPageTitle {
...
+
+  set title(title: string) {
+    this._title = title;
+    if (title !== '') {
+      title = `${title} | `;
+    }
+    this.bodyTitle.setTitle(`${title}Angular Material`);
+  }
+
+  constructor(private bodyTitle: Title) { }
```
jelbourn: Just noticed this breaks the unit test.

amcdnl: Good catch, fixed.

Issue #167, opened 2017-06-07
Add table-of-contents for jump navigation on docs


jelbourn:
Similar to the right-hand-side of https://developers.google.com/web/fundamentals/getting-started/principles/

amcdnl:
See: https://github.com/angular/material.angular.io/pull/230 and https://github.com/angular/material2/pull/6040

amysorto:
Closing this issue since a TOC for a page been a part of the docs site for a while now

PR #230, opened 2017-07-26
feat(toc): add table of contents to overview and guides

amcdnl:
Addresses #167 by adding a table of contents and jump links to the component overview and guide pages.

This PR depends on https://github.com/angular/material2/pull/6040 being merged.

img


src/app/shared/table-of-contents/table-of-contents.ts
```diff
@@ -0,0 +1,121 @@
...
+import 'rxjs/add/observable/fromEvent';
+import 'rxjs/add/operator/debounceTime';
+
+@Component({
+  selector: 'table-of-contents',
+  styleUrls: ['./table-of-contents.scss'],
+  template: `
+    <div *ngIf="links?.length" class="toc-container">
+      <h2>Contents</h2>
+      <ul>
```
jelbourn: `ul` and `ol` aren't actually the right elements for a list of navigation links; from an a11y standpoint lists are supposed to only contain text content. Better would just be a `<nav>` containing the anchors directly. If you need intermediate elements for styling you can use a `<div>` with `role="presentation"`
(I believe developers.google.com isn't doing the right thing here)

src/app/shared/table-of-contents/table-of-contents.ts
```diff
@@ -0,0 +1,121 @@
...
+            {{link.name}}
+          </a>
+        </li>
+      </ul>
+    </div>
+  `
+})
+export class TableOfContents implements OnInit, AfterViewInit {
+
+  @Input() links: any[] = [];
```
jelbourn: Can this be something other than `any`? Using it in this manner is pretty strongly discouraged 
(here and elsewhere)

src/app/shared/table-of-contents/table-of-contents.ts
```diff
@@ -0,0 +1,121 @@
...
+
+    this._scrollSubscription = Observable
+      .fromEvent(this._scrollContainer, 'scroll')
+      .debounceTime(10)
+      .subscribe(this.onScroll.bind(this));
+  }
+
+  ngAfterViewInit(): void {
+    this._route.fragment.subscribe(fragment => {
+      setTimeout(() => {
```
jelbourn: Why the 500ms timeout? At the very least it need an explanation in a comment, but better would be to not have it at all.
(here and below)

src/app/shared/table-of-contents/table-of-contents.ts
```diff
@@ -0,0 +1,121 @@
...
+    }, 500);
+  }
+
+  onScroll(): void {
+    for (let i = 0; i < this.links.length; i++) {
+      this.links[i].active = this.getLinkActive(this.links[i], this.links[i + 1]);
+    }
+  }
+
+  getLinkActive(currentLink: any, nextLink: any): boolean {
```
jelbourn: `isLinkActive`

src/app/shared/table-of-contents/table-of-contents.ts
```diff
@@ -0,0 +1,121 @@
...
+  }
+
+  ngOnInit(): void {
+    this._scrollContainer = this.container ?
+      this._document.querySelectorAll(this.container)[0] : window;
+
+    this._scrollSubscription = Observable
+      .fromEvent(this._scrollContainer, 'scroll')
+      .debounceTime(10)
+      .subscribe(this.onScroll.bind(this));
```
jelbourn: Need to unsubscribe on destroy

src/app/shared/table-of-contents/table-of-contents.ts
```diff
@@ -0,0 +1,121 @@
...
+  getLinkActive(currentLink: any, nextLink: any): boolean {
+    if (this.scrollOffset >= currentLink.top) {
+      if (nextLink && nextLink.top < this.scrollOffset) {
+        return false;
+      }
+      return true;
+    }
+    return false;
+  }
+
```
jelbourn:
```ts
return this.scrollOffset >= currentLink.top && !(nextLink && nextLink.top < this.scrollOffset);
```

Also with a comment like `A link is considered "active" if the page is scrolled passed it without also being scrolled passed the following link`

src/styles/_markdown-theme.scss
```diff
@@ -17,5 +17,25 @@
     background: rgba(mat-color($foreground, secondary-text), .01);
     border: .5px solid rgba(mat-color($foreground, secondary-text), .03);
   }
+
+  .docs-markdown-h3,
+  .docs-markdown-h4 {
+    position: relative;
+    padding-left: 30px;
+    margin-left: -30px;
```
jelbourn: Why the 30 / -30 here? At the very least needs a comment

src/styles/_markdown-theme.scss
```diff
@@ -17,5 +17,25 @@
...
+
+    .material-icons {
+      display: none;
+    }
+
+    &:hover {
+      .material-icons {
```
jelbourn: Looks like this is meant to target the header link but is more broad than that. Also why does it need to be absolute?

src/app/shared/table-of-contents/table-of-contents.ts
```diff
@@ -0,0 +1,121 @@
...
+import {Subscription} from 'rxjs/Subscription';
+import 'rxjs/add/observable/fromEvent';
+import 'rxjs/add/operator/debounceTime';
+
+@Component({
+  selector: 'table-of-contents',
+  styleUrls: ['./table-of-contents.scss'],
+  template: `
+    <div *ngIf="links?.length" class="toc-container">
+      <h2>Contents</h2>
```
jelbourn: I believe `<h2>` isn't appropriate here; `h1` - `h6` are meant for content section headings, not really navigation

src/app/pages/component-viewer/component-viewer.scss
```diff
@@ -24,7 +24,32 @@ app-component-viewer {
 }
 
 .docs-component-viewer-content {
+  position: relative;
   min-height: 500px;
+
+  table-of-contents {
```
amcdnl: Yes, it needs to override the host `top` position.

src/app/shared/header-link/header-link.ts
```diff
@@ -0,0 +1,30 @@
+import {Component, Input} from '@angular/core';
+import {Router} from '@angular/router';
+
+@Component({
+  selector: 'header-link',
+  template: `
+    <a
+      title="Link to this heading"
+      class="header-link docs-markdown-a"
+      aria-hidden="true"
```
amcdnl: I made it hidden because the Angular.io docs site had it this way.

src/app/shared/table-of-contents/table-of-contents.ts
amcdnl: I did notice MDN recommends `nav` + `ul` for navigation: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/nav

Funny W3 Schools recommends what you said: https://www.w3schools.com/tags/tag_nav.asp

src/app/shared/table-of-contents/table-of-contents.ts
amcdnl: Completely agree. Needed a mechanism to delay the jump so the content injection can load and bootstrap the components. Open to better ideas, I'll add a comment in the meantime.

src/styles/_markdown-theme.scss
amcdnl: Want the anchor to position 30px to the left of the header. Using absolutely position makes the anchor link not clickable because its outside the header hover.

Added comments for this.

src/styles/_markdown-theme.scss
amcdnl: See above on re: absolute.

amcdnl:
@jelbourn - ready for another review.

src/app/shared/table-of-contents/table-of-contents.ts
jelbourn: I feel so strange saying this, but I believe the MDN example is wrong and the w3schools is right. [The WAI-ARIA spec specifically says that the list role is for _non-interactive_ content](https://www.w3.org/TR/wai-aria/roles#list).

src/app/shared/doc-viewer/header-link.ts
```diff
@@ -0,0 +1,44 @@
...
+ */
+@Component({
+  selector: 'header-link',
+  template: `
+    <a
+      title="Link to this heading"
+      class="docs-markdown-a docs-header-link"
+      aria-label="Link to this heading"
+      [href]="url">
+      <i class="material-icons">link</i>
```
jelbourn: This `<i>` should be `aria-hidden` (or use `md-icon`)

src/app/shared/doc-viewer/header-link.ts
```diff
@@ -0,0 +1,44 @@
...
+ *
+ * would result in the wrong url, this component
+ * combines the current route with that jump link:
+ *
+ *    <a href="/guide#foo">Foo</a>
+ */
+@Component({
+  selector: 'header-link',
+  template: `
+    <a
```
jelbourn: I think this anchor would also benefit from an `aria-describedby` pointing to the header

src/app/shared/table-of-contents/table-of-contents.ts
```diff
@@ -0,0 +1,129 @@
...
+  Component, Input, AfterViewInit, Inject, ElementRef, OnInit
+} from '@angular/core';
+import {DOCUMENT} from '@angular/platform-browser';
+import {Router, ActivatedRoute, Event, NavigationEnd} from '@angular/router';
+import {Observable} from 'rxjs/Observable';
+import {Subscription} from 'rxjs/Subscription';
+import 'rxjs/add/observable/fromEvent';
+import 'rxjs/add/operator/debounceTime';
+
+interface Link {
```
jelbourn: Add some descriptions on this type? Not immediately obvious exactly what `type`, `top` etc. refer to or how `Link` in general is meant to be used

src/app/shared/table-of-contents/table-of-contents.ts
```diff
@@ -0,0 +1,129 @@
...
+  }
+
+  ngOnDestroy(): void {
+    this._routeSubscription.unsubscribe();
+    this._scrollSubscription.unsubscribe();
+  }
+
+  createLinks(): void {
+    // content is loading async, the timeout helps delay the query selector being empty
+    setTimeout(() => {
```
jelbourn: Would using `cdkObserveContent` / a `MutationObserver` be better for this?
(just to see when the content is updated and then can be unsubscribed)

src/app/shared/table-of-contents/table-of-contents.ts
```diff
@@ -0,0 +1,129 @@
...
+      const headers = this._document.querySelectorAll('.docs-markdown-h3,.docs-markdown-h4');
+      for (const header of headers) {
+        const { top } = header.getBoundingClientRect();
```
jelbourn: Omit spaces inside braces:
```ts
const {top} = header.getBoundingClientRect();
```
(here and elsewhere)

amcdnl:
@jelbourn - ready for second review.

jelbourn:
@amcdnl could you rebase?

amcdnl:
Done :)

src/app/shared/table-of-contents/table-of-contents.ts
```diff
@@ -0,0 +1,177 @@
...
+      <div class="docs-toc-heading">Contents</div>
+      <nav>
+        <a [href]="_rootUrl + '#' + link.id"
+          *ngFor="let link of links; let i = index"
+          class="docs-level-{{link.type}}"
+          [class.active]="link.active">
+          {{link.name}}
+        </a>
+      </nav>
+    </div>
```
jelbourn: Move this template into its own file?

src/app/shared/table-of-contents/table-of-contents.ts
```diff
@@ -0,0 +1,177 @@
...
+  @Input() headerSelectors = '.docs-markdown-h3,.docs-markdown-h4';
+
+  _rootUrl: string;
+  private _scrollContainer: any;
+  private _scrollSubscription: Subscription;
+  private _routeSubscription: Subscription;
+  private _fragmentObserver: MutationObserver;
+  private _headerObserver: MutationObserver;
+
+  private get scrollOffset(): number {
```
jelbourn: This would be a better function than a `getter`. 
(getters generate a lot of JS code vs just a function)

src/app/shared/table-of-contents/table-of-contents.ts
```diff
@@ -0,0 +1,177 @@
...
+    const {top} = this._element.nativeElement.getBoundingClientRect();
```
jelbourn: Add comment explaining what this is computing ("scrollOffset" by itself isn't specific enough)

src/app/shared/table-of-contents/table-of-contents.ts
```diff
@@ -0,0 +1,177 @@
...
+          this.createLinks();
+          this._rootUrl = rootUrl;
+        }
+      }
+    });
+  }
+
+  ngOnInit(): void {
+    this._scrollContainer = this.container ?
+      this._document.querySelectorAll(this.container)[0] : window;
```
jelbourn: Why `window` instead of `document.body` or `documentElement`?

src/app/shared/table-of-contents/table-of-contents.ts
```diff
@@ -0,0 +1,177 @@
...
+  @Input() links: Link[] = [];
+  @Input() container: string;
+  @Input() headerSelectors = '.docs-markdown-h3,.docs-markdown-h4';
+
+  _rootUrl: string;
+  private _scrollContainer: any;
+  private _scrollSubscription: Subscription;
+  private _routeSubscription: Subscription;
+  private _fragmentObserver: MutationObserver;
+  private _headerObserver: MutationObserver;
```
jelbourn: Add descriptions for these two observers?

src/app/shared/table-of-contents/table-of-contents.ts
```diff
@@ -0,0 +1,177 @@
...
+      this._fragmentObserver = new MutationObserver(() => {
+        const target = document.getElementById(fragment);
+        if (target) {
+          target.scrollIntoView();
+          this._fragmentObserver.disconnect();
+          this._fragmentObserver = null;
+        }
+      });
+
+      this._fragmentObserver.observe(document.body, {
```
jelbourn: Why attach this directly to the body instead of the document viewer host element?

src/app/shared/table-of-contents/table-of-contents.ts
```diff
@@ -0,0 +1,177 @@
...
+
+  createLinks(): void {
+    // ensure we don't create memory links
+    if (this._headerObserver) {
+      this._headerObserver.disconnect();
+    }
+
+    // observe body content changes until the headers
+    // are painted. this is required because the page can load async
+    this._headerObserver = new MutationObserver(() => {
```
jelbourn: I don't quite follow why you need two observers set up at different times

src/app/shared/table-of-contents/table-of-contents.ts
amcdnl: No reason, updated.

amcdnl: the body element doesn't have a height set img

src/app/shared/table-of-contents/table-of-contents.ts
amcdnl: This is attached to the body because it could be the document viewer or the guide viewer. In an effort, to keep it generic i attached it to the body. I could add an input if you like to pick a specific element to listen on. Let me know either way.

src/app/shared/table-of-contents/table-of-contents.ts
amcdnl: Unfortunately the router events do not give me enough information just to attach one to the route event.

For example, I listen to `this._route.fragment` which returns me the route fragment a user clicked. From that fragment I then ensure its loaded and then scroll into view.

On the `this._router.events`, I'm listening for when the user navigates to a totally different page. For example, navigating from accordion to autocomplete. In this scenario, the page doesn't actually change since its the same route it just updates the content. In this scenario, I want to update the links only.

If the router events gave me more information, I could easily consolidate these but at this time I need 2 different subscriptions and observers to make the appropriate action.

src/app/shared/table-of-contents/table-of-contents.ts
jelbourn: Both the component pages (`component-overview` and `component-api`) and the `guide-viewer` use `doc-viewer` in their template. Thinking about it more, though, I realize a much more straightforward approach would be to have the `doc-viewer` fire an event after appending the content, which the `table-of-contents` could use to know when to go, e.g.
```html
<doc-viewer
    documentUrl="/assets/documents/overview/{{componentViewer.componentDocItem.id}}.html"
    class="docs-component-view-text-content docs-component-overview"
    (contentLoaded)="toc.doThatThing()">
</doc-viewer>
<table-of-contents #toc
  container=".mat-sidenav-content">
</table-of-contents>
```
I must have had `MutationObserver` on my mind the other day

amcdnl: Lets talk through this, I'm curious to your thoughts on integrating the route change and the observer together.

src/app/shared/table-of-contents/table-of-contents.ts
```diff
@@ -0,0 +1,135 @@
...
+  private _scrollContainer: any;
+  private _scrollSubscription: Subscription;
+  private _routeSubscription: Subscription;
+  private _fragmentSubscription: Subscription;
+  private _urlFragment = '';
+
+  constructor(private _router: Router,
+              private _route: ActivatedRoute,
+              private _element: ElementRef,
+              @Inject(DOCUMENT) private _document: any) {
```
jelbourn: Any reason the type here can't be `Document`?

src/app/shared/table-of-contents/_table-of-contents-theme.scss
```diff
@@ -0,0 +1,20 @@
...
+  $foreground: map-get($theme, foreground);
+
+  .docs-toc-container {
+    border-left: solid 4px mat-color($primary);
+
+    a {
+      color: mat-color($foreground, secondary-text);
+
+      &:hover,
+      &.active {
```
jelbourn: Need to prefix these classes with `docs-`

src/app/shared/table-of-contents/table-of-contents.scss
```diff
@@ -0,0 +1,29 @@
...
+}
+
+.docs-toc-heading {
+  margin: 0;
+  padding: 0;
+  font-size: 13px;
+  font-weight: bold;
+}
+
+a {
```
jelbourn: Change you make this style a css class instead of styling `a` directly?

package.json
```diff
@@ -4,7 +4,7 @@
   "license": "MIT",
   "angular-cli": {},
   "scripts": {
-    "start": "npm run build-themes && ng serve --aot --sourcemaps=false",
+    "start": "npm run build-themes && ng serve --sourcemaps=false",
```
amcdnl: I didn't mean to commit this but AOT causes lots of issues when developing and is slow.

amcdnl:
@jelbourn - good to go.

Issue #232, opened 2017-07-29
docs(code): highlight inline code element


donroyco:
Proposal.

Right now, code elements which are inside a paragraph look very similar to the non-code elements. This is due to (almost) the same font-size and font-family of the code elements and the text. 

What if we change the inline code elements' `background-color`, add some `padding` and give them a small `border` to make the text more readable (CSS is taken from the `.docs-markdown-pre` class with some added padding).

*Current look*
![current-look](https://user-images.githubusercontent.com/1763537/28742691-198a4be0-7437-11e7-8cab-b0f2a22608a9.png)

*New look*
![new-look](https://user-images.githubusercontent.com/1763537/28742696-488b6820-7437-11e7-829f-f013ebef0b15.png)

It's much easier to distinct the code from non-code. Of course this is just a proposal which can be altered, but I think it will benefit people if they want to quickly scan the page.

amcdnl:
@donroyco - I believe I addressed this in https://github.com/angular/material.angular.io/pull/234

josephperrott:
Fixed by #234

PR #234, opened 2017-07-30
docs(code): highlight inline code elements #232

amcdnl:
Highlights inline code examples in the overview and api docs.

**Light**
light

**Dark**
dark

Addresses: https://github.com/angular/material.angular.io/issues/232


src/styles/_api-theme.scss
```diff
@@ -23,4 +23,16 @@
   .docs-api-h1 {
     display: none !important;
   }
+
+  .docs-api-class-name,
+  .docs-api-module-import,
+  .docs-api-class-selector-name,
+  .docs-api-class-export-name {
```
jelbourn: Can do
```scss
$exportBackgroundOpacity: if($is-dark-theme, 0.06, 0.03);
background: rgba(mat-color($foreground, secondary-text), $exportBackgroundOpacity);
```

amcdnl: updated :)

Issue #168, opened 2017-06-07
Support multiple versions of documentation


jelbourn:

TheRedD0t:
Any progress? Now that beta versions are also released, it makes even more sense to have a possibility to choose between the different versions.

jelbourn:
Done in v6.

chrste90:
@jelbourn Do you also plan to add a next version to the docs like on `angular.io`?

jelbourn:
At some point, but not in the immediate future

PR #246, opened 2017-08-31
feat: Add the version number to the page.

josephperrott:


jelbourn:
I don't really like the way this looks. How about we put it in the footer?

amcdnl:
@jelbourn - Where would you like this to be in the footer?  Maybe here img

jelbourn:
I was thinking in the empty space left of the copywrite

willshowell:
Pinging, since the latest release has demoed the usefulness of this.

jelbourn:
@josephperrott rebase?

josephperrott:
@jelbourn rebased.  Though I never updated to change the placement of the versioning.

I have concerns for the placing it in the footer on mobile, I think it will be hard to discover as the footer looks busier.

willshowell:
It should probably go where a multiple-version select would be (https://github.com/angular/material.angular.io/issues/168), right? Maybe in the sidenav like [material.angularjs.org](https://material.angularjs.org)?

TheRedD0t:
I like the design with a select of the versions like at [material.angularjs.org](https://material.angularjs.org), too! Why reinvent the wheel?

willshowell:
Any updates on this? Would hate to see it missed for the next release.

jelbourn:
We'll have something to switch between versions eventually. For now I think we can start in the footer and explore from there

ericruiz-developer:
So will these conflicts be addressed and merged or is this not happening any longer? Version would be really help IMO.

Issue #247, opened 2017-09-01
I cannot read the component document on mobile, I can't scroll


tom10271:
As titles with my iPhone 6s Plus in iOS 10.3 and latest version of Chrome

willshowell:
This also occurs in iOS Safari and macOS Safari

amcdnl:
Looks like its being caused by the drawers `mat-drawer-content` div.

tom10271:
I have revisited the doc with my iPhone. I can scroll now but the font size is quite large, probably too large on mobile

PR #248, opened 2017-09-01
bug(docs): fix scroll broken on safari #247

amcdnl:
Addresses #247 where scrolling doesn't work on safari.


willshowell:
@amcdnl Changing the selector to `.mat-drawer-content` here should do it:

https://github.com/angular/material.angular.io/blob/047b7334f87f29ce397088e852a4a2e7c6f9039f/src/app/pages/component-sidenav/component-sidenav.scss#L8-L16

amcdnl:
@jelbourn @willshowell - Ya, that works better. Updated.

Issue #221, opened 2017-07-12
Add a "copy" button in examples


rafaelss95:
Add a button to make it easier to copy examples, like in https://angular.io:

![](https://i.stack.imgur.com/81h53.jpg)

PR #262, opened 2017-09-08
feat(example-viewer): add copy text button

willshowell:
- This basically copies the feature from core/aio 
- I made the button hide/show on hover since it was distracting while switching between tabs

Fixes https://github.com/angular/material.angular.io/issues/221

### DIFF
--- a/src/app/shared/copier/copier.service.ts
+++ b/src/app/shared/copier/copier.service.ts
@@ -10,58 +10,51 @@ import {Injectable} from '@angular/core';
 
 @Injectable()
 export class CopierService {
-  private fakeElem: HTMLTextAreaElement;
+
+  private textarea: HTMLTextAreaElement;
+
+  /** Copy the text value to the clipboard. */
+  copyText(text: string): boolean {
+    this.createTextareaAndSelect(text);
+
+    const copySuccessful = document.execCommand('copy');
+    this.removeFake();
+
+    return copySuccessful;
+  }
 
   /**
-   * Creates a fake textarea element, sets its value from `text` property,
+   * Creates a hidden textarea element, sets its value from `text` property,
    * and makes a selection on it.
    */
-  createFake(text: string) {
-    const isRTL = document.documentElement.getAttribute('dir') === 'rtl';
-
+  private createTextareaAndSelect(text: string) {
     // Create a fake element to hold the contents to copy
-    this.fakeElem = document.createElement('textarea');
+    this.textarea = document.createElement('textarea');
 
     // Prevent zooming on iOS
-    this.fakeElem.style.fontSize = '12pt';
+    this.textarea.style.fontSize = '12pt';
 
-    // Reset box model
-    this.fakeElem.style.border = '0';
-    this.fakeElem.style.padding = '0';
-    this.fakeElem.style.margin = '0';
-
-    // Move element out of screen horizontally
-    this.fakeElem.style.position = 'absolute';
-    this.fakeElem.style[ isRTL ? 'right' : 'left' ] = '-9999px';
+    // Hide the element
+    this.textarea.classList.add('cdk-visually-hidden');
 
     // Move element to the same position vertically
     const yPosition = window.pageYOffset || document.documentElement.scrollTop;
-    this.fakeElem.style.top = yPosition + 'px';
+    this.textarea.style.top = yPosition + 'px';
 
-    this.fakeElem.setAttribute('readonly', '');
-    this.fakeElem.value = text;
+    this.textarea.setAttribute('readonly', '');
+    this.textarea.value = text;
 
-    document.body.appendChild(this.fakeElem);
+    document.body.appendChild(this.textarea);
 
-    this.fakeElem.select();
-    this.fakeElem.setSelectionRange(0, this.fakeElem.value.length);
+    this.textarea.select();
+    this.textarea.setSelectionRange(0, this.textarea.value.length);
   }
 
-  removeFake() {
-    if (this.fakeElem) {
-      document.body.removeChild(this.fakeElem);
-      this.fakeElem = null;
-    }
-  }
-
-  copyText(text: string) {
-    try {
-      this.createFake(text);
-      return document.execCommand('copy');
-    } catch (err) {
-      return false;
-    } finally {
-      this.removeFake();
+  /** Remove the text area from the DOM. */
+  private removeFake() {
+    if (this.textarea) {
+      document.body.removeChild(this.textarea);
+      this.textarea = null;
     }
   }
 }
--- a/src/app/shared/copier/copier.service.ts
+++ b/src/app/shared/copier/copier.service.ts
@@ -1,0 +1,67 @@
+/**
+ * This class is based on the code in the following projects:
+ *
+ * - https://github.com/zenorocha/select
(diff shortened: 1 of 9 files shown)


src/app/shared/example-viewer/example-viewer.ts
```diff
@@ -44,4 +50,12 @@ export class ExampleViewer {
   exampleFileUrl(extension: string) {
     return `/assets/examples/${this.example}-example-${extension.toLowerCase()}.html`;
   }
+
+  copySource(text: string) {
+    if (this.copier.copyText(text)) {
+      this.snackbar.open('Code Copied', '', {duration: 800});
```
donroyco: I would make this `'Code copied'`, lowercase second `c`.

willshowell: Agreed!

willshowell:
cc @jelbourn

src/app/shared/example-viewer/example-viewer.html
```diff
@@ -19,8 +19,15 @@
     <md-tab-group>
       <!-- TODO(jelbourn): don't hard-code the html + ts + css structure -->
       <md-tab *ngFor="let extension of ['HTML', 'TS', 'CSS']" [label]="extension">
-        <pre *ngIf="exampleFileUrl(extension)" class="docs-example-source"><doc-viewer
-            [documentUrl]="exampleFileUrl(extension)"></doc-viewer></pre>
+        <div class="docs-example-source-wrapper">
+          <button md-icon-button type="button" class="docs-example-source-copy"
```
josephperrott: Can you have the attributes on the next line start at the same column

```
<button md-icon-button type="button" class="docs-example-source-copy"
        title="Copy code snippet" (click)="copySource(view.textContent)">
```

src/app/shared/example-viewer/example-viewer.scss
```diff
@@ -21,8 +21,21 @@
   flex: 1 1 auto;
 }
 
+.docs-example-source-copy {
```
josephperrott: Can you add a rule for `dir: rtl` to place it on the left?

src/app/shared/example-viewer/example-viewer.ts
```diff
@@ -22,6 +23,11 @@ export class ExampleViewer {
   /** Whether the source for the example is being displayed. */
   showSource = false;
 
+  constructor(
```
josephperrott: Collapse the white space at the end of the constructor definition

```
constructor(
  private snackbar: MdSnackBar,
  private copier: CopierService) {}
```

src/app/shared/example-viewer/example-viewer.ts
```diff
@@ -44,4 +50,12 @@ export class ExampleViewer {
   exampleFileUrl(extension: string) {
     return `/assets/examples/${this.example}-example-${extension.toLowerCase()}.html`;
   }
+
+  copySource(text: string) {
+    if (this.copier.copyText(text)) {
+      this.snackbar.open('Code copied', '', {duration: 1000});
```
josephperrott: I think that `2500ms` is probably a better duration, since `1000ms` essentially is enough time to appear and then immediately hide.

src/app/shared/copier/copier.service.ts
```diff
@@ -0,0 +1,67 @@
...
+ * - https://github.com/zenorocha/select
+ * - https://github.com/zenorocha/clipboard.js/
+ *
+ * Both released under MIT license - © Zeno Rocha
+ */
+import {Injectable} from '@angular/core';
+
+@Injectable()
+export class CopierService {
+  private fakeElem: HTMLTextAreaElement;
```
jelbourn: `fakeElem` -> `textarea`?

It's weird to call it fake, since it's a real `textarea`, just hidden

src/app/shared/copier/copier.service.ts
```diff
@@ -0,0 +1,67 @@
...
+
+  /**
+   * Creates a fake textarea element, sets its value from `text` property,
+   * and makes a selection on it.
+   */
+  createFake(text: string) {
```
jelbourn: `createTextareaForCopy`? Add JsDoc description?

src/app/shared/copier/copier.service.ts
```diff
@@ -0,0 +1,67 @@
...
+    const isRTL = document.documentElement.getAttribute('dir') === 'rtl';
+
+    // Create a fake element to hold the contents to copy
+    this.fakeElem = document.createElement('textarea');
+
+    // Prevent zooming on iOS
+    this.fakeElem.style.fontSize = '12pt';
+
+    // Reset box model
+    this.fakeElem.style.border = '0';
```
jelbourn: Would it work to just add `cdk-visually-hidden` to the element?

src/app/shared/copier/copier.service.ts
```diff
@@ -0,0 +1,67 @@
...
+    this.fakeElem.setAttribute('readonly', '');
+    this.fakeElem.value = text;
+
+    document.body.appendChild(this.fakeElem);
+
+    this.fakeElem.select();
+    this.fakeElem.setSelectionRange(0, this.fakeElem.value.length);
+  }
+
+  removeFake() {
```
jelbourn: Do this in an `ngOnDestroy`?

src/app/shared/copier/copier.service.ts
```diff
@@ -0,0 +1,67 @@
...
+    if (this.fakeElem) {
+      document.body.removeChild(this.fakeElem);
+      this.fakeElem = null;
+    }
+  }
+
+  copyText(text: string) {
+    try {
```
jelbourn: Under what circumstances would this throw an error?

https://github.com/angular/material2/blob/master/CODING_STANDARDS.md#try-catch

willshowell: This service is straight from [core](https://github.com/angular/angular/blob/master/aio/src/app/shared/copier.service.ts). It looks like the try/catch originally comes from [here](https://github.com/zenorocha/clipboard.js/commit/231d62db0e64e8eaeea3bf3b6627418f1ced23ab), but as far as I can tell, that was written before `execCommand` was supported by all major browsers. Will remove.

src/app/shared/copier/copier.service.ts
willshowell: I'm not sure I follow. Since this service is provided at the module level, it won't be destroyed, right? I can call this from the ExampleViewer's ngOnDestroy, but I don't see an advantage to that. Seems like it's best to clean up the dom as soon as the copying is complete?

jelbourn: Nevermind, I was thinking the `textarea` stuck around between copies

src/app/shared/copier/copier.service.ts
jelbourn: I have higher code quality standards than core :wink:

src/app/shared/example-viewer/example-viewer.html
```diff
@@ -19,8 +19,14 @@
...
```
jelbourn: Last thing I noticed: needs an `aria-label` on the button for something like "Copy example source to clipboard"

(`title` unfortunately doesn't work well as a label for some screen readers)

Issue #236, opened 2017-08-09
Proposal: stop wrapping code samples


willshowell:
Wrapping for the code examples was added in #95, but I can't find any conversations about it and would like to just suggest not wrapping them. Personally, I think the x-scroll behavior on Github is nice,

```scss
pre {
  // Pre elements won't wrap text by default and can exceed the boundaries of their parent element.
  // This adds extra scroll space when visiting the docs mobile. Pre-wrapping the code examples
  // will ensure that the code never exceeds the page width.
  white-space: pre-wrap;
}
```

For some examples, the wrapping makes the code impossible to read:

![screen shot 2017-08-09 at 12 13 43 pm](https://user-images.githubusercontent.com/6475576/29131900-53063dc0-7cfc-11e7-85d7-5a0a459ee069.png)

whereas without wrapping, you can read the code formatted in the way it was intended:

![screen shot 2017-08-09 at 12 15 13 pm](https://user-images.githubusercontent.com/6475576/29131935-79f84b6c-7cfc-11e7-8841-549810803536.png)

jelbourn:
FYI, unless someone is mentioned on the PR in this repo it will go unnoticed for a while.

josephperrott:
I agree that we need to prevent wrapping the code snippets, however it is a bit more complicated as the styling make it a bit more complex to wrap the code snippet, without causing the entire page to scroll on the x axis.  Will have to sort it out.

PR #265, opened 2017-09-12
fix: Make site more mobile friendly

josephperrott:
- Make navbar "sticky" on all pages
- Show table of contents (non-sticky) at the top of pages on mobile
- Shrink landing page spacing on mobile to increase density
- Move navbar to a two row version on mobile
- Use md-icon for angular and github icons in navbar
- Enforce word breaks for p tags in docs, preventing x axis overflows
- Decrease height of Guides subheader to match other pages
- Shrink docs-api table spacing on mobile to increase density

### DIFF
--- a/src/app/material-docs-app.html
+++ b/src/app/material-docs-app.html
@@ -1,2 +1,2 @@
-<app-navbar [class.mat-elevation-z6]="showShadow"></app-navbar>
+<app-navbar class="mat-elevation-z6"></app-navbar>
 <router-outlet></router-outlet>
--- a/src/app/material-docs-app.scss
+++ b/src/app/material-docs-app.scss
@@ -12,6 +12,11 @@ material-docs-app > app-component-sidenav {
   flex: 1 1 auto;
 }
 
+// Target last-child as it will be the output of the router-outlet within the app's container.
+material-docs-app > :last-child {
+  overflow-y: auto;
+}
+
 app-navbar {
   position: relative;
   z-index: 10;
--- a/src/app/shared/navbar/navbar.html
+++ b/src/app/shared/navbar/navbar.html
@@ -1,19 +1,30 @@
 <!-- TODO: figure out if the <nav> should go inside of a <header> element. -->
-<nav class="docs-navbar">
+<nav class="docs-navbar-header">
   <a md-button class="docs-button" routerLink="/" aria-label="Angular Material">
     <img class="docs-angular-logo"
-      src="../../../assets/img/homepage/angular-white-transparent.svg"
-      alt="angular">
+         src="../../../assets/img/homepage/angular-white-transparent.svg"
+         alt="angular">
     <span>Material</span>
   </a>
-  <a md-button class="docs-button" routerLink="components">Components</a>
-  <a md-button class="docs-button" routerLink="guides">Guides</a>
+  <a md-button class="docs-navbar-hide-small docs-button" routerLink="components">Components</a>
+  <a md-button class="docs-navbar-hide-small docs-button" routerLink="guides">Guides</a>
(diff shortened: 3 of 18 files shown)


src/app/shared/navbar/navbar.html
```diff
@@ -1,19 +1,22 @@
 <!-- TODO: figure out if the <nav> should go inside of a <header> element. -->
-<nav class="docs-navbar">
+<nav class="docs-header">
   <a md-button class="docs-button" routerLink="/" aria-label="Angular Material">
-    <img class="docs-angular-logo"
-      src="../../../assets/img/homepage/angular-white-transparent.svg"
-      alt="angular">
+    <md-icon class="angular-logo" svgIcon="angular"></md-icon>
```
jelbourn: Any reason to make this change? Switch to `md-icon` does incur a cost, and we don't need to have an SVG here since we don't style the Angular logo

(same for github logo)

src/app/material-docs-app.scss
```diff
@@ -12,6 +12,10 @@ material-docs-app > app-component-sidenav {
   flex: 1 1 auto;
 }
 
+material-docs-app > :last-child {
```
jelbourn: Add a comment that describes what you're targeting here?

src/app/shared/navbar/navbar.scss
```diff
@@ -15,14 +15,42 @@
...
-  vertical-align: middle;
+.angular-logo {
+  height: 30px;
+  width: 30px;
+}
+
+.docs-navbar {
+  display: none;
+
+  .docs-nav-link {
```
jelbourn: Can this rule be un-nested?

src/app/shared/navbar/navbar.scss
```diff
@@ -15,14 +15,42 @@
...
+    text-decoration: inherit;
+    flex: 1;
+  }
+}
+
+.show-small {
```
jelbourn: Missing prefix (here and elsewhere). The prefix is typically something like `docs-navbar-`

src/styles/_api-theme.scss
```diff
@@ -25,6 +25,11 @@
     display: none !important;
   }
 
+  // Prevent p tags from not breaking, causing x axis overflows.
+  .docs-api > p {
+    word-break: break-word;
+  }
```
jelbourn: Why direct child?

src/app/material-docs-app.html
```diff
@@ -1,2 +1,2 @@
-<app-navbar [class.mat-elevation-z6]="showShadow"></app-navbar>
+<app-navbar class="mat-elevation-z6"></app-navbar>
```
willshowell: Any way to only make the elevation show once the page has been scrolled and stickiness has taken effect? If not, I think `showShadow` can be refactored out of the component class.

willshowell:
Also please consider https://github.com/angular/material.angular.io/issues/236! The code sample wrapping makes it really hard to read on mobile!

src/app/material-docs-app.html
josephperrott: I don't believe its worth the effort to make the elevation only when scrolled, especially since we are faking the stickiness of the header.  I have removed `showShadow` as you mentioned.

src/app/shared/navbar/navbar.html
jelbourn: I think it should be okay to keep using `img`; I want to get better about encouraging people to only use `md-icon` when there'd be some advantage in doing so

jelbourn:
Looks like a unit test is failing

josephperrott:
@jelbourn unit test issue resolved

PR #267, opened 2017-09-19
fix: set initial focus point for pages

josephperrott:


src/app/pages/component-viewer/component-viewer.ts
```diff
@@ -38,8 +38,14 @@ export class ComponentViewer {
...
   encapsulation: ViewEncapsulation.None,
 })
-export class ComponentOverview {
+export class ComponentOverview implements OnInit {
+  @ViewChild('intialFocusTarget') focusTarget: ElementRef;
+
   constructor(public componentViewer: ComponentViewer) {}
+
+  ngOnInit() {
+    window.setTimeout(() => this.focusTarget.nativeElement.focus(), 100);
```
jelbourn: Shouldn't need `window.`

jelbourn: Add comment that explains the 100ms delay?

src/app/shared/navigation-focus/navigation-focus.ts
```diff
@@ -0,0 +1,40 @@
...
+
+@Injectable()
+export class Focuser {
+  focusTimeout: number;
+
+  focus(el: HTMLElement) {
+    window.clearTimeout(this.focusTimeout);
+    this.focusTimeout = window.setTimeout(() => el.focus(), 100);
+  }
+}
```
jelbourn: I don't think you need the separate `Focuser` service. We can store the last timeout id in the js module since we know we only ever want there to be one.

```ts
let lastTimeoutId = -1;

@Directive({
  selector: '[focusOnNavigation]',
  host: {'tabindex': '-1'}
})
export class NavigationFocus implements OnInit {
  constructor(private elementRef: ElementRef) {}

  ngOnInit() {
    clearTimeout(lastTimeoutId);
    lastTimeoutId = setTimeout(() => elementRef.nativeElement.focus(), 100);
  }
}
```
(also baked in a few style suggestions there)

josephperrott:
@jelbourn Comments addressed.

PR #289, opened 2017-10-13
fix(example-viewer): don't wrap code samples

willshowell:
Fixes https://github.com/angular/material.angular.io/issues/236


willshowell:
cc @jelbourn for review

PR #290, opened 2017-10-14
fix: fix components and CDK categories page title

futurizing:
This PR fixes the bug on where the title and the primary header is currently not set when nagivating to https://material.angular.io/components/categories and https://material.angular.io/cdk/categories.


futurizing:
Looks like the bug has already been fixed with the new redesign.

PR #300, opened 2017-10-19
feat(nav): new sidenav

amcdnl:
Implements new sidenav.

img

img

img

PR #310, opened 2017-11-01
New sidenav

amcdnl:
This PR implements the new left nav. This is the exact same PR as #300 just merged w/ @mmalerba latest changes.


mmalerba:
I see a couple of issues here:
- In mobile mode, the document does not receive scroll events (this is needed for the address bar to scroll up on phones (compare to https://material2-docs-dev.firebaseapp.com/components/autocomplete/overview where the document does receive scroll events)
- In mobile mode, there seems to be a gap between the 2 toolbars
docs-site1
- In desktop mode, there is a gap at the bottom (below the footer)
docs-site2
- In desktop mode, when I expand the sidenav both the content and the sidenav scroll (the content doesn't actually need to scroll, it's only scrolling because of the sidenav) 
docs-site3

amcdnl:
1) So that is tricky because of the sticky positioned items. I was able to work around this by adding a hack that will automatically hide the browser address bar for you.
2) Fixed
3) Fixed
4) I'm not sure its possible to fix this given the nature of things that need to scroll / push in this scenario. Open to suggestions though.

amcdnl:
tag @jelbourn

jelbourn:
@mmalerba can you review this?

mmalerba:
Sorry, I thought this wasn't ready since it has conflicts. @amcdnl Can you resolve it so I can bring up a demo?

amcdnl:
@mmalerba - merged :)

amcdnl:
@mmalerba - I believe I had addressed all your feedback and am ready for review! :)

src/app/pages/component-sidenav/_component-sidenav-theme.scss
```diff
@@ -1,36 +1,49 @@
...
-  $accent: map-get($theme, accent);
-  $warn: map-get($theme, warn);
   $background: map-get($theme, background);
   $foreground: map-get($theme, foreground);
-  $sidenav: '.docs-component-viewer-sidenav';
+  $is-dark-theme: map-get($theme, is-dark);
+  $nav-background-opacity: if($is-dark-theme, 0.2, 0.03);
 
+  .docs-component-viewer-nav-content {
+    background: rgba(0, 0, 0, $nav-background-opacity);
```
mmalerba: should this be the foreground color instead of always black?

src/app/pages/component-sidenav/_component-sidenav-theme.scss
```diff
@@ -1,36 +1,49 @@
...
 
-  .docs-component-viewer-sidenav {
+    &::-webkit-scrollbar-thumb {
+      background: rgba(0, 0, 0, .26);
```
mmalerba: same comment as above

src/app/pages/component-sidenav/_component-sidenav-theme.scss
```diff
@@ -1,36 +1,49 @@
...
+  @media (max-width: $small-breakpoint-width) {
+    .docs-component-viewer-sidenav {
+      .docs-component-viewer-nav-content {
+        background: none;
+      }
+    }
+
+    .mat-drawer {
+      &::-webkit-scrollbar-thumb {
+        background: rgba(0, 0, 0, .26);
```
mmalerba: same comment as above

src/app/pages/component-sidenav/component-nav.html
```diff
@@ -0,0 +1,23 @@
...
+          <a [routerLink]="'/' + (params | async)?.section+ '/' + component.id"
+            routerLinkActive="docs-component-viewer-sidenav-item-selected">
+            {{component.name}}
+          </a>
+        </li>
+      </ul>
+      <hr *ngIf="!last" />
+    </nav>
+  </div>
+</div>
```
mmalerba: nit: \n

src/app/pages/component-sidenav/component-sidenav.html
```diff
@@ -1,25 +1,25 @@
...
+          *ngIf="!isScreenSmall()"
+          [params]="params">
+        </app-component-nav>
+        <router-outlet></router-outlet>
+      </div>
+      <app-footer></app-footer>
+    </div>
   </div>
-</mat-sidenav-container>
+</mat-sidenav-container>
```
mmalerba: nit: \n

mmalerba:
Awesome, it looks much better now. It looks like this one is still an issue:

> - In desktop mode, when I expand the sidenav both the content and the sidenav scroll (the content doesn't actually need to scroll, it's only scrolling because of the sidenav) 
docs-site3

I don't think it's a huge deal, I would be ok submitting the PR if it's too difficult to fix, but it would be nice to fix.

amcdnl:
@mmalerba - The nav is always a fixed 75vh so it should be scrolling in this case. The body scrolling in this case is unexpected and I thought I had fixed this. Let me take another look and let you know.

src/app/pages/component-sidenav/_component-sidenav-theme.scss
amcdnl: I wanted to make it a slight lighter accent of the current background. This is mainly done like this for different themes.

amcdnl:
@mmalerba - Really the only good fix for this would be to remove the scrollbar from the side nav and just let it push down. I implemented the scrollbar like this similar to the google api docs that I used as an example. Let me know how you would like to proceed.

mmalerba:
we can't do something like this?

```css
max-height: calc(100vh - <header-height> - <footer-height>)
```

PR #316, opened 2017-11-03
fix: consistent toc positioning

willshowell:
1. Shares the TOC styles for the component-overview and component-api components
2. Refactors the styles for the guide viewer to make the toc sticky and position on top for small screens

Note: Making the toc sticky on the guide viewer illuminates that it doesn't really work (live highlighting of current section), but I think that would be fixed by using cdkScrollable, which is a change I'd also like to make.


willshowell:
cc @jelbourn

jelbourn:
LGTM

PR #321, opened 2017-11-03
refactor(toc): use cdkScrollable for scrolling, remove unnecessary inputs

willshowell:
- Combine duplicated logic
- (try to) increase readability
- Use cdkScrollable for subscribing to scroll events
- Use `_activeLinkIndex` to ensure only a single link is active at a time (removes `active` from link interface)
- Remove input decorator from `links`
- Fix bug where navigating to a page with a fragment would not scroll the header into view until clicking on the link

Note: this still does not work for toc on the guides pages. I thought scrolling would be on the window and therefore be picked up by the scroll dispatcher, but that does not seem to be the case.

cc @jelbourn


src/app/shared/table-of-contents/table-of-contents.ts
```diff
@@ -43,92 +39,84 @@ export class TableOfContents implements OnInit {
...
   }
 
-  updateScrollPosition(): void {
-    this.links = this.createLinks();
-
-    const target = document.getElementById(this._urlFragment);
-    if (target) {
-      target.scrollIntoView();
-    }
+  contentReady(): void {
```
jelbourn: I'd rather name this method for what it _does_ rather than when it's called

src/app/shared/table-of-contents/table-of-contents.ts
```diff
@@ -43,92 +39,84 @@ export class TableOfContents implements OnInit {
...
-  private getScrollOffset(): number {
-    const {top} = this._element.nativeElement.getBoundingClientRect();
-    if (typeof this._scrollContainer.scrollTop !== 'undefined') {
-      return this._scrollContainer.scrollTop + top;
-    } else if (typeof this._scrollContainer.pageYOffset !== 'undefined') {
-      return this._scrollContainer.pageYOffset + top;
-    }
+  /** Find the target from the url fragment and scroll it into view. */
+  private scrollFragmentIntoView(): void {
+    this._ngZone.runTask(() => {
```
jelbourn: Why does this need to enter the zone if it's only updating the scroll position?

src/app/shared/table-of-contents/table-of-contents.ts
```diff
@@ -43,92 +39,84 @@ export class TableOfContents implements OnInit {
...
+      const target = document.getElementById(this._urlFragment);
+      if (target) {
+        // scrollIntoView may require next macrotask if target has just been loaded
```
jelbourn: I don't quite follow why this is necessary; the `document-viewer` knows exactly when the content is loaded and in the DOM

src/app/shared/table-of-contents/table-of-contents.ts
willshowell: Oops that was leftover from debugging something else. Removed.

src/app/shared/table-of-contents/table-of-contents.ts
willshowell: Without it, scroll position would always end up at the top of the page (can be seen [here](https://material.angular.io/components/autocomplete/overview#setting-separate-control-and-display-values)). 

However I now recall that there was some other component that manually updates the scroll position. I'll try to track down any sort of race condition occurring there.

willshowell: Yeah it's conflicting with the navigation focus directive, https://github.com/angular/material.angular.io/blob/f42d5f412fef560e9891b23a89558c22f914f2f6/src/app/shared/navigation-focus/navigation-focus.ts#L14-L18

willshowell:
@jelbourn ready for another review. The component overview and component api components were autofocusing the visually hidden element at the top after 100ms. I've changed them to wait until the content has loaded, but before the TOC scrolls the correct header into view.

Would it be better for a11y to place focus on the header link when the fragment does exist?

jelbourn:
@willshowell Yeah, if there is a fragment then the autofocus should be on the header for that fragment. I think that was just an oversight when doing the autofocus. Would you be interested in addressing that in this PR?

willshowell:
@jelbourn Ugh I thought I'd be able to easily, but I'm hitting some strange behavior where `focus()` is noop on the header-link unless `setTimeout` is used. I'd like to defer that to a later PR.

jelbourn:
@amcdnl can you review?

src/app/shared/table-of-contents/table-of-contents.ts
```diff
@@ -1,11 +1,10 @@
...
+import {Component, ElementRef, Inject, Input, NgZone, OnDestroy, OnInit} from '@angular/core';
 import {DOCUMENT} from '@angular/platform-browser';
 import {ActivatedRoute, NavigationEnd, Router} from '@angular/router';
-import 'rxjs/add/observable/fromEvent';
-import 'rxjs/add/operator/debounceTime';
-import 'rxjs/add/operator/takeUntil';
-import {Observable} from 'rxjs/Observable';
+import {ScrollDispatcher, CdkScrollable} from '@angular/cdk/scrolling';
 import {Subject} from 'rxjs/Subject';
+import 'rxjs/add/operator/filter';
```
amcdnl: Lets use Rx Lettables here. I have a PR to transform the rest of them in the app.

willshowell:
@jelbourn @amcdnl rebased and comments addressed

PR #329, opened 2017-11-18
fix: create proper fragment urls

DevVersion:
By default fragment links are referring to the root page and the actual hash won't matter anymore. This can be fixed in a similar way to the `header-link` directive by updating the anchor URLs to include the current base URL.

Since those are normal links and aren't `header-links`, the `header-link` directive is not an option. A directive that updates the links accordingly would be perfect, but there is no way to interpolate directives on those HTML templates dynamically.

Having a component that creates/wraps the anchor elements is wrong and adds more pain than it helps. The links should can be just updated using simple DOM manipulation.

Fixes https://github.com/angular/material2/issues/8503


DevVersion:
cc. @jelbourn

PR #338, opened 2017-11-28
Fix unit test and lint failures

jelbourn:


jelbourn:
I think the docs assets are messing this up

PR #343, opened 2017-12-02
fix(examples): fix broken sidenav example

amcdnl:
Addresses https://github.com/angular/material2/issues/8682

### DIFF
--- a/src/app/shared/doc-viewer/doc-viewer.ts
+++ b/src/app/shared/doc-viewer/doc-viewer.ts
@@ -91,7 +91,7 @@ export class DocViewer implements OnDestroy {
       let portalHost = new DomPortalHost(
           element, this._componentFactoryResolver, this._appRef, this._injector);
       let examplePortal = new ComponentPortal(componentClass, this._viewContainerRef);
-      let exampleViewer = portalHost.attach(examplePortal);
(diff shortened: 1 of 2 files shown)


src/app/shared/doc-viewer/doc-viewer.ts
```diff
@@ -91,7 +91,7 @@ export class DocViewer implements OnDestroy {
       let portalHost = new DomPortalHost(
           element, this._componentFactoryResolver, this._appRef, this._injector);
       let examplePortal = new ComponentPortal(componentClass, this._viewContainerRef);
-      let exampleViewer = portalHost.attach(examplePortal);
+      let exampleViewer: any = portalHost.attach(examplePortal);
```
jelbourn: Why is this `any`? It should be able to infer the right `ComponentRef`, no?

amcdnl: I added `any` here to fix the error on the next line where it was attaching `example` to the instance and not being found. I forgot about this change, didn't mean to commit. I'll fix the type :).

amcdnl:
Looks like @mmalerba made a commit to fix this, closing this PR.

Ref: https://github.com/angular/material.angular.io/pull/346

PR #344, opened 2017-12-02
chore: switch to rxjs lettables

amcdnl:
This PR switches RXJS usage to lettable operators.


src/app/material-docs-app.ts
```diff
@@ -1,6 +1,6 @@
 import {Component, ViewEncapsulation} from '@angular/core';
 import {Router, NavigationEnd} from '@angular/router';
-import 'rxjs/add/operator/filter';
+import {filter} from 'rxjs/operators';
```
crisbeto: This should be imported from `rxjs/operators/filter`, otherwise it'll import all the operators. Same goes for all the other `rxjs/operators` imports.

amcdnl: Good call. Fixed.

amcdnl:
@jelbourn - fixed :)

jelbourn:
@amcdnl can you rebase?

amcdnl:
@jelbourn done :)

PR #352, opened 2017-12-11
fix(table): http example needs httpclientmodule

### LINKED ISSUE #339: HttpClientModule is missing in stackblitz MatTableModule examples
When you open stackblitz's example in the [example's tab](https://material.angular.io/components/table/examples) => "Table retrieving data through HTTP" example, `HttpClientModule` is missing, so the example doesn't run.

image

andrewseguin:
Current example is broken since it requires the new Http module.

Fixes #339

### DIFF
--- a/src/assets/stackblitz/main.ts
+++ b/src/assets/stackblitz/main.ts
@@ -39,6 +39,7 @@ import {
 } from '@angular/material';
 import {MaterialDocsExample} from './app/material-docs-example';
 import {HttpModule} from '@angular/http';
+import {HttpClientModule} from '@angular/common/http';
 import {CdkTableModule} from '@angular/cdk/table';
 
 @NgModule({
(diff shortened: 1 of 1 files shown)


mmalerba:
is there a corresponding fix in material2 repo that this is generated from?

andrewseguin:
Yeah this one here - https://github.com/angular/material2/commit/0ea43700efb105aa02ad3b1d3f800bf4a9925883#diff-c3b988af35d5ab4790e4795e2d7015dd

src/assets/stackblitz/main.ts
```diff
@@ -86,6 +87,7 @@ export class DemoMaterialModule {}
     BrowserAnimationsModule,
     FormsModule,
     HttpModule,
```
Splaktar: Should we remove `HttpModule`?

josephperrott: `HttpModule` will be removed everywhere in a different PR.

andrewseguin: Good call - let's cleanup in another PR. Would like to at least just make sure this fix goes in without the risk of breaking other examples if I miss a case of one using `HttpModule`

PR #363, opened 2018-01-01
chore: bump year

crisbeto:
Happy New Year! :tada:

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,6 +1,6 @@
 The MIT License
 
-Copyright (c) 2017 Google, Inc.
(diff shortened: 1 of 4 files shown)


rafaelss95:
Looks like there are more places that need this change:

`src/app/shared/footer/footer.html`
`src/app/shared/stackblitz/stackblitz-writer.spec.ts`
`src/app/shared/stackblitz/stackblitz-writer.ts` (this one, instead of using a hard coded year, I'd concatenate with `new Date().getFullYear()` :) )).

PR #372, opened 2018-01-23
Resolved duplicate class tag SVG error

SamuelMarks:
Here's where I found the error, plus some others to checkout:
screen shot 2018-01-23 at 4 33 08 pm

Resolved your fastandconsistent.svg error.


Chan4077:
Isn't this a duplicate of #361?

SamuelMarks:
Looks right, missed that one

Issue #373, opened 2018-01-23
Navigating to stackblitz examples should show component TS file, not main.ts


andrewseguin:
Most users will want to see an examples TS file for the _component_, not the _main.ts_

amcdnl:
Not sure on the API for this, cc/ @EricSimons

amcdnl:
Done #390

Splaktar:
Can confirm that this is done.

PR #390, opened 2018-02-10
feat(stackblitz): start sb on index #373

amcdnl:
This PR starts the stackblitz demo on the index file. Addresses #373


pjuchniewicz:
The test was not updated

PR #395, opened 2018-02-14
fix: markdown docs overlap

ekoeryanto:
The docs overlap the right sidenav as seen at angular/material2#9951


ekoeryanto:
I signed it!

PR #400, opened 2018-02-18
fix(stackblitz): fix svg path getting extra /app

amcdnl:
This PR fixes the SVG demo not getting the correct path. There is still an issue with the SVG not being resolved though. Reached out to Eric regarding.

See: https://stackblitz.com/edit/angular-niug11?file=app/icon-svg-example.ts

### DIFF
--- a/src/app/shared/stackblitz/stackblitz-writer.ts
+++ b/src/app/shared/stackblitz/stackblitz-writer.ts
@@ -114,7 +114,14 @@ export class StackblitzWriter {
     form.appendChild(input);
   }
 
-  /** Reads the file and adds its text to the form */
+  /**
+   * Reads the file and adds its text to the form
+   * @param form the html form you are appending to
+   * @param data example metadata about the example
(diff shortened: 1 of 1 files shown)


src/app/shared/stackblitz/stackblitz-writer.ts
```diff
@@ -115,9 +115,13 @@ export class StackblitzWriter {
   }
 
   /** Reads the file and adds its text to the form */
-  _readFile(form: HTMLFormElement, data: ExampleData, filename: string, path: string): void {
+  _readFile(form: HTMLFormElement,
+            data: ExampleData,
+            filename: string,
+            path: string,
+            prependApp = true): void {
```
jelbourn: Now would be a good time to add param descriptions, since `prependApp` is the first really non-obvious one

amcdnl: Added comments. :)

PR #411, opened 2018-03-14
fix(tests): fix tests on CI

josephperrott:
- Remove e2e tests which acted as no-op.
- Fix unit tests
- Update to Angular and Material 6 betas.


src/app/pages/component-viewer/component-viewer.spec.ts
```diff
@@ -17,6 +17,12 @@ const mockActivatedRoute = {
   fragment: Observable.create(observer => {
     observer.complete();
   }),
+  parent: {
+    params: Observable.create(observer => {
```
mmalerba: `observableOf({})`?

PR #424, opened 2018-04-03
chore: update site to latest 6.0.0 RCs

jelbourn:
@amcdnl here's my change that just updates everything. It works when given the docs assets for 6.0.0-rc.0

(the errors you were seeing were probably because support for native `<table>` with `MatTable` was added after rc.0, so the examples from master wouldn't work with that).


amcdnl:
This is no longer relevant

PR #431, opened 2018-04-26
Update to Angular CLI v6

jelbourn:


src/app/shared/example-viewer/example-viewer.spec.ts
```diff
@@ -51,30 +50,30 @@ describe('ExampleViewer', () => {
     component = fixture.componentInstance;
   });
 
-  it('should toggle showSource boolean', () => {
+  it('should toggle showSource boolean', async(() => {
```
josephperrott: Shouldn't we be doing `fakeAsync` here?

PR #503, opened 2018-08-15
fix: remove service worker

DevVersion:
* Removes the service worker for the material.angular.io docs page
* Unregisters previously registered service workers and force-reloads the page if someone opens the docs and still has a service worker installed.


troyspencer:
Why?

PR #512, opened 2018-08-27
feat: add summaries to all menu items and category pages

benelliott:
This PR adds short (max. two-line) summaries to all list cards across the documentation to provide more context to the user before they navigate to the documentation. A short paragraph is also added at the top of the two category list pages ("Components"/"CDK") to provide some context to those.

Currently I feel like there is not enough information on the documentation list pages to A) explain where you are and B) hint at where you want to go. This means that users could be sent a link like https://material.angular.io/cdk/categories and have the following questions:

- What is the CDK? What does it stand for? Why would I use it instead of "normal" Angular Material?
- What is in the "Components" menu item and how is that different from the "Components" button in the toolbar?

Alternatively, they may visit https://material.angular.io/components/categories and be unsure what to expect when they click one of the child items. Will "Layout" contain components for structuring pages at the top-level, such as a sidebar, or will that be in "Navigation"?

I was surprised to see while writing this PR that there used to be a summary field for the categories but this was removed in https://github.com/angular/material.angular.io/commit/5d7429db4eff63d03a252bad1be84c4188c6c89a. Was there a particular reason for this?

I don't expect that much of the content of these descriptions to make it through to the site unchanged but would love it to start a discussion about how to make the menu pages more contextualised.

### DIFF
--- a/src/app/pages/component-category-list/component-category-list.html
+++ b/src/app/pages/component-category-list/component-category-list.html
@@ -1,9 +1,11 @@
+<div class="docs-component-category-list-summary">{{_categoryListSummary}}</div>
 <div class="docs-component-category-list">
   <a *ngFor="let category of docItems.getCategories((params | async)?.section)"
      class="docs-component-category-list-item"
     [routerLink]="['../', category.id]">
     <mat-card class="docs-component-category-list-card">
       <mat-card-title>{{category.name}}</mat-card-title>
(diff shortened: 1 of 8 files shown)


jelbourn:
This is great! It helps quite a lot. Looks good aside from needing a rebase; I may tweak some of the summaries in a follow-up PR.

benelliott:
@jelbourn Great, thanks! Rebased and added summaries for text-field and platform.

PR #514, opened 2018-09-02
docs: show text-field, coercion and platform in CDK docs

DevVersion:
Not sure whether coercion is something we want to show though. Also the `coerceArray` description feels a bit unclear to me.


src/app/shared/documentation-items/documentation-items.ts
```diff
@@ -244,11 +244,14 @@ const DOCS: {[key: string]: DocCategory[]} = {
       items: [
         {id: 'a11y', name: 'Accessibility', examples: []},
         {id: 'bidi', name: 'Bidirectionality', examples: []},
+        {id: 'coercion', name: 'Coercion', examples: []},
```
jelbourn: Lets leave `coercion` out for now; I'd like to revisit it if/when we add more stuff

DevVersion: Done.

PR #516, opened 2018-09-08
feat(example-viewer): support rendering additional files

DevVersion:
* Currently we only render the `HTML`, `TS` and `CSS` files in the view-source tab group. This is fine for most examples, but there are also examples like the `form-field-custom-control` that prefer to have the template of the "custom control" externally. We should support rendering these additional files in the tab group instead of just magically adding them to the external StackBlitz that will be generated.

Related to angular/material2#13021


jelbourn:
Can you add a screenshot of what this looks like for an example with multiple assets?

DevVersion:
image

Is that screenshot sufficient? There is also another tab for the `my-tel-input-example.css` file (that is invisible here)

PR #518, opened 2018-09-12
build: fix continuous integration

DevVersion:


DevVersion:
@jelbourn This moves away from Saucelabs and Browserstack and just runs against Firefox and Chrome Headless because maintaining the SL and BS tooling is kind of a pain I guess.

This makes sure that the CI runs better in long-term without active maintaining. The CI failed for **a long** time on this repository.

PR #519, opened 2018-09-12
fix: anchor links not scrolling into view

### LINKED ISSUE #396: section links don't work until after page load
**Steps:**
Try navigating from off-site directly to a section header:
https://material.angular.io/components/input/overview#changing-when-error-messages-are-shown

**Result:**
Page loads scrolled to top.

**Expected:**
Page loads scrolled to the header I linked to.

**Notes:**
Once the page is loaded I can click header links and they work as expected.

DevVersion:
* Currently if someone visits a link that contains a fragment that refers to an anchor element, the element won't be scrolled into view properly. This is because the `focusOnNavigation` calls `focus()` after the scroll into view. This causes the focused element to be visible in view.
* Removes an unnecessary workaround for fixing fragment URLs.

Fixes #396

### DIFF
--- a/src/app/pages/component-viewer/component-viewer.ts
+++ b/src/app/pages/component-viewer/component-viewer.ts
@@ -65,7 +65,7 @@ export class ComponentViewer implements OnDestroy {
   encapsulation: ViewEncapsulation.None,
 })
 export class ComponentOverview implements OnInit {
-  @ViewChild('intialFocusTarget') focusTarget: ElementRef;
+  @ViewChild('initialFocusTarget') focusTarget: ElementRef;
   @ViewChild('toc') tableOfContents: TableOfContents;
   showToc: Observable<boolean>;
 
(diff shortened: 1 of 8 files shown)


src/app/pages/component-viewer/component-viewer.ts
```diff
@@ -76,10 +76,10 @@ export class ComponentOverview implements OnInit {
 
   ngOnInit() {
     // 100ms timeout is used to allow the page to settle before moving focus for screen readers.
-    setTimeout(() => this.focusTarget.nativeElement.focus(), 100);
+    setTimeout(() => this.focusTarget.nativeElement.focus({preventScroll: true}), 100);
   }
 
-  onContentLoaded() {
+  onContentRendered() {
```
jelbourn: Could we call this `scrollToSelectedContentSection"? In general it's better to name functions for what they do rather than how they're called.

jelbourn:
Is the CI failure just from needing rebase?

DevVersion:
@jelbourn Yeah, this PR has been created before the CI fixes have been merged. Rebased now.

src/app/pages/component-viewer/component-viewer.ts
DevVersion: Definitely. I've just tried to follow the naming conventions that existed before. Changed now.

PR #524, opened 2018-09-24
Added dropdown icon to version picker

anagi202:
A simple, but important change -

Added an arrow dropdown icon to the version picker. In my opinion, it is need immediately obvious that you can change versions. (I wasted a lot of time looking for the A5 website the other day). It doesn't really look interactive, it looks like it's just telling me the most recent version. 

A dropdown icon will make it more apparent that you can switch between A5 and A6 documentation. 

screen shot 2018-09-24 at 12 58 57 am


anagi202:
I signed the CLA!

PR #538, opened 2018-10-18
chore: update 5.x branch to include new versions

josephperrott:


josephperrott:
confirmed cla as this conflict was caused by a cherry-pick

PR #548, opened 2018-10-20
fix(ci): hammerjs not loaded causes warnings and flaky failures

Splaktar:
import hammerjs for the tests
shorten a few imports

I didn't import hammerjs in `main.ts` since that isn't needed to fix the tests, but it does seem like it should be there. Was there a reason to avoid enabling it on the docs site? Should I add it in?

### DIFF
--- a/package.json
+++ b/package.json
@@ -33,6 +33,7 @@
     "@angular/platform-browser-dynamic": "^7.0.0",
     "@angular/router": "^7.0.0",
     "core-js": "^2.5.6",
+    "hammerjs": "^2.0.8",
     "moment": "^2.22.1",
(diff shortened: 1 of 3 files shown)


Splaktar:
You can see that the test runs are much cleaner in this one:
https://travis-ci.org/angular/material.angular.io/jobs/443947111

Than this one with warnings:
https://travis-ci.org/angular/material.angular.io/jobs/443944450

I don't have an example of the flaky failures atm, but a few builds have needed to be restarted lately due to them.

DevVersion:
@Splaktar Needs a rebase.

Splaktar:
@DevVersion thanks for the heads up. Rebased.

PR #558, opened 2018-11-22
Properly show extends clause.

DevVersion:
Combined with: https://github.com/angular/material2/pull/14232, thisfixes that the extends clause anchor will be shifted by the `header-link` styling.

Also improves the general visuals of the extends clause, since those are now clickable. Additionally updates the lock file to reflect the latest package changes.

### DIFF
--- a/src/styles/_api.scss
+++ b/src/styles/_api.scss
@@ -69,6 +69,14 @@
   padding: 3px;
 }
 
+.docs-api-class-extends-type {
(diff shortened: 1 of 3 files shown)


DevVersion:
@jelbourn Since #14232 has been merged, it would be good to merge this one as well. Otherwise if we forget about this PR for the next release, extends clauses would look _very_ bad.

PR #565, opened 2018-12-22
Use docs-content from NPM package

devversion:
Blocked on updating the publish job on Material, so that it only publishes the NPM package.


devversion:
@jelbourn This one is ready for review. The changes in `material2` have been merge, so that one is last remaining PR related to that.

PR #567, opened 2019-01-02
build: update to Angular 7.1

crisbeto:
Bumps to the latest minor Angular version.


devversion:
No longer needed. Closing since we are on V8 already. https://github.com/angular/material.angular.io/commit/13c76b1c2e7ff90c49a1a45d97e39a9368652e7a

PR #583, opened 2019-02-23
docs(readme): update development instructions

### LINKED ISSUE #578: docs(readme): instructions to build and contribute to repo are out of date
README.md has the following instruction:
```
Build and copy docs and examples from Angular Material: npm run fetch-local
```
However, `package.json` has no `fetch-local` script.

swseverance:
* remove development setup instructions that are no longer relevant
* resolves #578


benjaminmaldoner:
I still get the following error (even though I'm using your updated dev instructions):
ERROR in : Unexpected value 'ExampleModule in C:/git/material.angular.io/node_modules/@angular/material-examples/example-module.d.ts' imported by the module 'AppModule in C:/git/material.angular.io/src/app/app-module.ts'. Please add a @NgModule annotation.
How do I solve it?

swseverance:
@benjaminmaldoner I'm able to replicate the error you're seeing but I don't know how to fix it.

jelbourn:
It's currently a bug in the bazel rule we use to build the examples package. We're waiting for a fix from the Angular tooling team.

README.md
```diff
@@ -24,15 +24,8 @@ The documentation is generated from the following resources
 - [CDK components, services, and directives](https://github.com/angular/material2/tree/master/src/cdk)
 
 ## Development Setup
-1. Clone Angular Material in the parent directory of this repo
-    1. `cd ..`
-    1. `git clone git@github.com:angular/material2.git`
 1. Install Gulp globally: `npm i -g gulp`
```
laertispappas: Shouldn't this line be removed as well? Do we still need gulp?

MatthiasKunnen: Gulp is no longer used so it should be removed too.

README.md
```diff
@@ -25,8 +25,7 @@ The documentation is generated from the following resources
 - [CDK components, services, and directives](https://github.com/angular/components/tree/master/src/cdk)
 
 ## Development Setup
-1. Install Gulp globally: `npm i -g gulp`
-2. Run `npm install`
+1. Run `npm install`
```
jelbourn: We're actually using `yarn` here now if you want to change it to that

swseverance: `README.md` refers to `npm` commands in several other places, so for the sake of consistency I would prefer to have `npm install`. If you would prefer to replace all usage of `npm` commands with their `yarn` equivalents I can close this PR and leave that task to someone else.

PR #587, opened 2019-03-13
chore: run explicit static-query timing migration

devversion:
This is the result of the static-query ng-update schematic. Mostly ran it for testing, but should be good to keep if we update the app to V8-beta.


devversion:
cc. @mmalerba this could be merged now since we upgraded to V8

PR #590, opened 2019-04-03
refactor: run v8 material import schematic

devversion:


src/app/shared/stackblitz/stackblitz-writer.ts
```diff
@@ -1,6 +1,6 @@
 import {HttpClient} from '@angular/common/http';
 import {Injectable} from '@angular/core';
-import {VERSION} from '@angular/material';
+import {VERSION} from '@angular/material/typings';
```
devversion: This is an interesting issue that came up. Importing from there is invalid. This will be solved if we move the `VERSION` to `core`, but for now this breaks. cc. @jelbourn

devversion: This is solved with https://github.com/angular/components/pull/15700

PR #601, opened 2019-05-27
fix(manifest.json): use rel paths, remove density

### LINKED ISSUE #599: Manifest icons don't work propery
This links in your `manifest.json`:
```
"src": "\/android-chrome-36x36.png",
        ^^
```
should be more like this:
```
"src": "./android-chrome-36x36.png",
        ^^
```
because this breaks the thing in Firefox.

See https://github.com/webcompat/web-bugs/issues/30891

marcoscaceres:
fixes #599
 
The paths are broken and the escapes "\" are unnecessary. Also, there is no longer `density` member in the spec. It was removed a few years ago.

### DIFF
--- a/src/assets/img/favicons/manifest.json
+++ b/src/assets/img/favicons/manifest.json
@@ -8,40 +8,34 @@
   "orientation": "portrait",
   "icons": [
     {
-      "src": "\/android-chrome-36x36.png",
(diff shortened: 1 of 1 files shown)


Splaktar:
Verified this locally. It fixed some favicon 404s that I was seeing.

PR #634, opened 2019-09-20
build: fix object might be undefined error in template

Splaktar:
Relates to #631


Splaktar:
Merging to fix this breaking local dev builds.

PR #641, opened 2019-10-02
refactor: convert app to latest Angular CLI format

Splaktar:
## PR Checklist

Please check that your PR fulfills the following requirements:
- [x] The commit message follows [our guidelines](https://github.com/angular/components/blob/master/CONTRIBUTING.md#commit)
- [x] Tests for the changes have been added or weren't appropriate

## What is the current behavior?
The Angular workspace and app project structure is not up to date with the latest recommendations from the Angular CLI.

Issue Number: 
N/A

## What is the new behavior?
Update to the latest recommendations from the Angular CLI in order to facilitate a future update to enable Ivy.

- fix some Ivy and TypeScript compilation and configuration issues
- add a `build:highmem` for use when the build fails due to memory usage
- update all dependencies that are supported
- remove unneeded polyfills now that differential loading is supported
- convert workspace to current format
- move and update tsconfigs to match current CLI format
- log errors on bootstrap
- enable some router quality of life options
- add a `Version` interface
- update StackBlitz
  - Copyright
  - major Angular version
  - dependencies
- fix build error for `setTimeout`
  - it was resolving to the NodeJS version rather than DOM version
- Work around https://github.com/angular/components/issues/17264

## Other information
This fixes some Ivy-related issues, but there are a few remaining fixes needed here and in Angular Material itself before the project will build with Ivy.


angular.json
```diff
@@ -1,10 +1,16 @@
 {
+  "$schema": "./node_modules/@angular/cli/lib/config/schema.json",
```
jelbourn: Are the changes for Angular CLI structure throughout this PR captured by a schematic? To what extend will these changes be required of other Angular CLI users? (Particularly ones who started a project in the early days like us)

cc @vikerman @IgorMinar @StephenFluin

e2e/src/app.po.ts
```diff
@@ -0,0 +1,12 @@
+import {browser, by, element} from 'protractor';
+
+export class MaterialDocsAppPage {
```
jelbourn: Mark these functions `async`?

e2e/src/app.e2e-spec.ts
```diff
@@ -0,0 +1,25 @@
...
+describe('Material Docs App', () => {
+  let page: MaterialDocsAppPage;
+
+  beforeEach(() => {
+    page = new MaterialDocsAppPage();
+  });
+
+  it('should display welcome message', () => {
+    page.navigateTo();
+    page.getTitleText().then((title: string) => {
```
jelbourn: Use `await` instead of `then`?

package.json
```diff
@@ -4,60 +4,65 @@
...
     "start-jit": "yarn build-themes && ng serve --aot=false",
-    "lint": "tslint -p src/tsconfig.json",
+    "start:prod": "yarn build-themes && ng serve --prod",
+    "lint": "tslint -p tsconfig.json",
     "test": "yarn build-themes && ng test",
     "pree2e": "webdriver-manager update",
-    "e2e": "protractor",
+    "e2e": "yarn build-themes && ng e2e",
+    "build": "yarn build-themes && ng build",
+    "build:highmem": "yarn build-themes && node --max_old_space_size=8192 ./node_modules/@angular/cli/bin/ng build",
```
jelbourn: Is this something commonly required, or is there something special about our app? I'm surprised that it's necessary given the relatively simple nature of our docs app

package.json
```diff
@@ -4,60 +4,65 @@
...
-    "core-js": "^2.6.1",
+    "@angular/animations": "^8.2.8",
+    "@angular/cdk": "^8.2.2",
+    "@angular/cdk-experimental": "^8.2.2",
+    "@angular/common": "^8.2.8",
+    "@angular/compiler": "^8.2.8",
+    "@angular/core": "^8.2.8",
+    "@angular/forms": "^8.2.8",
+    "@angular/material": "^8.2.2",
+    "@angular/material-examples": "angular/material2-docs-content#518c443",
```
jelbourn: Could this be `#master` until we release v9?

src/app/pages/component-sidenav/component-sidenav.spec.ts
```diff
@@ -15,17 +16,17 @@ describe('ComponentSidenav', () => {
...
   it('should close the sidenav on init', () => {
-    const component = fixture.componentInstance;
-
     // Spy on window.mediaMatch and return stub
     spyOn(window, 'matchMedia').and.returnValue({
       matches: true
     } as any);
 
-    fixture.detectChanges();
+    // TODO refactor this as none of these expectations are ever verified
```
jelbourn: What do you mean by "verified"?

package.json
Splaktar: It's tracked here: https://github.com/angular/angular-cli/issues/13734. Based on this issue, it has become quite common and only minor progress has been made towards "solving it". In some cases, the CLI team has stated that increasing the available memory is expected.

I only ran into the issue when doing an AOT non-production build.

package.json
Splaktar: This commit sha maps to the latest `8.2.x` snapshot release prior to `8.2.2-0462599` which yarn won't install due to invalid semver. Reported in: https://github.com/angular/components/issues/17264 and fixed in PR https://github.com/angular/components/pull/17267.

Just leaving it as `angular/material2-docs-content` is equivalent to `#master` AFAIK and that does indeed build just fine.

I'm not sure whether or not all of the docs changes for `8.2.x` are getting merged down to `master` though?

src/app/pages/component-sidenav/component-sidenav.spec.ts
Splaktar: Currently this test generates the following warning in the console:
```
'Spec 'ComponentSidenav should close the sidenav on init' has no expectations.'
```
I debugged it quite a bit, but I wasn't able to figure out how to re-write the test in a way that
1. Actually ran the `expect()` calls
1. Was configured in a way that `component.sidenav` was defined

e2e/src/app.po.ts
Splaktar: I can do that, but it doesn't line up with what the Angular CLI generates.

Should we open an issue to update Angular CLI's e2e generation schematic to always make these `async` functions?

src/app/shared/navigation-focus/navigation-focus.ts
```diff
@@ -13,7 +13,8 @@ export class NavigationFocus implements OnInit {
   ngOnInit() {
     clearTimeout(lastTimeoutId);
     // 100ms timeout is used to allow the page to settle before moving focus for screen readers.
-    lastTimeoutId = setTimeout(() => this.el.nativeElement.focus({preventScroll: true}), 100);
+    lastTimeoutId = window.setTimeout(() => this.el.nativeElement.focus({preventScroll: true}),
```
Splaktar: For some reason the build always failed here because it thought this was referring to NodeJS's function which returns a `NodeJS.Timeout` or `NodeJS.Timer`. Adding `window` was the widely accepted solution across StackOverflow and issue trackers.

src/app/pages/component-sidenav/component-sidenav.html
```diff
@@ -4,7 +4,7 @@
               *ngIf="(isScreenSmall | async)"
               [opened]="!(isScreenSmall | async)"
               [mode]="(isScreenSmall | async) ? 'over' : 'side'"
-              [fixedInViewport]="(isScreenSmall | async)"
+              [fixedInViewport]="(isScreenSmall | async) ? true : false"
```
Splaktar: This is required with Ivy enabled due to
```ts
component-sidenav.html(7,15): Type 'boolean | null' is not assignable to type 'boolean | undefined'.
```
In `MatSidenav` it appears to have the type of `'boolean | undefined'`
```ts
@Input()
get fixedInViewport(): boolean { return this._fixedInViewport; }
set fixedInViewport(value) { this._fixedInViewport = coerceBooleanProperty(value); }
private _fixedInViewport = false;
```
From `boolean-property.ts`:
```ts
export function coerceBooleanProperty(value: any): boolean {
  return value != null && `${value}` !== 'false';
}
```
In `AsyncPipe`, the type is `'boolean | null'`
```ts
transform<T>(obj: Observable<T> | null | undefined): T | null;
```
@jelbourn do you think that our boolean `@Input()`s and `coerceBooleanProperty()` need to be updated to `'boolean | null'` instead of `'boolean'` (which is implicitly `'boolean | undefined'`)?

src/app/shared/theme-picker/theme-picker.ts
```diff
@@ -71,8 +71,8 @@ export class ThemePicker implements OnInit, OnDestroy {
 
   ngOnInit() {
     this._queryParamSubscription = this._activatedRoute.queryParamMap
-      .pipe(map(params => params.get('theme')), filter(Boolean))
-      .subscribe(themeName => this.installTheme(themeName));
+      .pipe(map((params: ParamMap) => params.get('theme')), filter(Boolean))
+      .subscribe((themeName: string) => this.installTheme(themeName));
```
Splaktar: This is needed for `strictNullChecks`, but it wasn't detected previously. With the latest Angular/CLI/RxJS versions, it does become an error:
```ts
theme-picker.ts:75:51 - error TS2345: Argument of type 'unknown' is not assignable to parameter of type 'string'.
```

src/app/shared/version-picker/version-picker.ts
```diff
@@ -13,10 +13,10 @@ const versionUrl = 'https://material.angular.io/assets/versions.json';
   templateUrl: './version-picker.html'
 })
 export class VersionPicker {
-  /** The currently running versin of Material. */
+  /** The currently running version of Material. */
   materialVersion = materialVersion;
   /** The possible versions of the doc site. */
-  docVersions = this.http.get(versionUrl);
+  docVersions = this.http.get<Array<VersionInfo>>(versionUrl);
```
Splaktar: I've added types here, but this still fails in Ivy. I plan to address the Ivy failure in another branch.
```ts
version-picker.html(6,25): Type 'VersionInfo[] | null' is not assignable to type 'VersionInfo[] | Iterable<VersionInfo> | undefined'.
version-picker.html(8,8): Object is of type 'unknown'.
```

src/main.ts
```diff
@@ -9,10 +9,11 @@ import {unregisterServiceWorkers} from './unregister-service-workers';
 // Unregister all installed service workers and force reload the page if there was
 // an old service worker from a previous version of the docs.
 unregisterServiceWorkers()
-  .then(hadServiceWorker => hadServiceWorker && location.reload(true));
+  .then(hadServiceWorker => hadServiceWorker && location.reload());
```
Splaktar: Passing `true` for `forceReload` is deprecated.

package.json
Splaktar: I'm not sure if this would occur with Ivy as well since the Ivy build doesn't get that far into the build process, yet.

e2e/src/app.po.ts
jelbourn: It would be worth starting a discussion at least

package.json
jelbourn: We merge everything into master and then cherry-pick into `8.2.x` (except odd/exceptional cases)

src/app/shared/navigation-focus/navigation-focus.ts
jelbourn: cc @devversion the nodejs typings shouldn't be making their way into the web compilation unit, but this is something we can deal with later

src/app/pages/component-sidenav/component-sidenav.html
jelbourn: For now this is fine; I've previously proposed a set of features to the framework team that would solve this more ergonomically, but until then we'll just deal with the annoying behavior

angular.json
Splaktar: Most of the changes are not breaking or required changes. They mostly fall into the category of best practices. Over time, as these slight deviations from best practices build up, you can run into issues and/or unexpected breaking changes in future migrations. 

Most projects that I've seen will try to align themselves with the best practices (i.e. output from a `ng new`) from the latest version of the Angular CLI every couple of major releases. Currently, this repo seems to be aligned with the recommendations from Angular CLI v5.

jelbourn:
I just created an `8.x` branch on this repo for the current master; we can publish any remaining 8.x releases from there and make target v9 with master

angular.json
jelbourn: I guess what I really want to know here is how much of this change was for ivy and how much was was just event horizon for Angular CLI?

Splaktar: I just went through and added comments for any Ivy-specific issues.

e2e/src/app.e2e-spec.ts
```diff
@@ -0,0 +1,23 @@
...
+
+  it('should display welcome message', async() => {
+    await page.navigateTo();
+    expect(await page.getTitleText()).toEqual('Angular Material');
+  });
+
+  afterEach(async () => {
+    // Assert that there are no errors emitted from the browser
+    const logs = await browser.manage().logs().get(logging.Type.BROWSER);
+    expect(logs).not.toContain(jasmine.objectContaining({
```
crisbeto: I'm not sure whether this will actually fail the test. A long time ago I tried doing something similar to detect leaking elements in unit tests and I couldn't get it to fail properly, although Jasmine could've changed the behavior since then. It might be better to either throw an error here or pull it out into a function that is called at the end of each test.

src/app/shared/version-picker/version-picker.ts
crisbeto: AFAIK we usually prefer `VersionInfo[]` instead of `Array<VersionInfo>`.

tsconfig.app.json
```diff
@@ -0,0 +1,17 @@
+{
+  "extends": "./tsconfig.json",
+  "compilerOptions": {
+    "outDir": "./out-tsc/app",
+    "types": ["q", "selenium-webdriver", "jasmine"]
```
crisbeto: Is `q` actually used anywhere?

e2e/src/app.e2e-spec.ts
Splaktar: This is taken directly from what the Angular CLI's e2e schematic generates for a new project.

If there is a little more concrete guidance here, I could open up an issue against Angular CLI.

src/app/shared/version-picker/version-picker.ts
Splaktar: Updated.

package.json
Splaktar: It's also needed with Ivy.

e2e/src/app.po.ts
Splaktar: PR https://github.com/angular/angular-cli/pull/15785 proposes this change in `@schematics/angular`.

package.json
Splaktar: It looks like this is being addressed in PR https://github.com/angular/angular-cli/pull/15777.

e2e/src/app.po.ts
Splaktar: It looks like using `async` functions and `await` breaks e2e tests on IE11, Edge, Safari, and Firefox. It also breaks on Chrome 41 and 75.

PR #642, opened 2019-10-03
fix: prepare for enabling Ivy

Splaktar:
## PR Checklist

Please check that your PR fulfills the following requirements:
- [x] The commit message follows [our guidelines](https://github.com/angular/components/blob/master/CONTRIBUTING.md#commit)
- [x] Tests for the changes have been added or weren't appropriate

## What is the current behavior?
- The app doesn't build with Ivy enabled due to https://github.com/angular/angular/pull/31371
- StackBlitz examples
  - load the deprecated `@angular/http` library
  - include out of date polyfills
  - include an out of date `.angular-cli.json`

Issue Number: 
N/A

## What is the new behavior?
- remove use of AsyncPipe with `*ngFor`
  - due to https://github.com/angular/angular/pull/31371
- remove deprecated `@angular/http` from StackBlitz dependencies
- remove out of date `.angular-cli.json` from StackBlitz templates
  - it's unclear that moving to the new `angular.json` format is viable
- remove unneeded polyfills from StackBlitz templates

## Other information
There is one remaining issue blocking the app from building with Ivy:
`_MatTabLinkBase` needs the `@Directive()` treatment i.e.
```ts
@Directive({
  // TODO(crisbeto): this selector can be removed when we update to Angular 9.0.
  selector: 'do-not-use-abstract-mat-tab-link-base'
}
```
@crisbeto is fixing it in https://github.com/angular/components/pull/17282.


src/app/shared/version-picker/version-picker.html
```diff
@@ -3,7 +3,7 @@
   <mat-icon>arrow_drop_down</mat-icon>
 </button>
 <mat-menu #versionPicker="matMenu">
-  <button mat-menu-item *ngFor="let version of docVersions | async"
+  <button mat-menu-item *ngFor="let version of docVersions"
```
Splaktar: AsyncPipe removed due to the interaction of types defined for NgForOf and AsyncPipe (tracked in https://github.com/angular/angular/pull/31371).

Alternatives were to type it as `Observable<any>` or disable one of `strictNullChecks`, `enableIvy`, or `fullTemplateTypeCheck`.
Reproduction and more details at: https://github.com/Splaktar/ngfor-asyncpipe-ivy.

Splaktar:
Going to verify the StackBlitz changes on the dev instance.

I also got another idea to try for allowing the use of the AsyncPipe with `*ngFor`.

src/app/shared/version-picker/version-picker.html
Splaktar: It looks like this workaround may work:
https://github.com/angular/angular/pull/32980/files?file-filters%5B%5D=.html#diff-986a7d69debab319598e3a2e92ea615dR24

Splaktar: and it does. Will update in a follow up PR.

PR #651, opened 2019-10-13
fix(docs): correct overflow cdk pages

donroyco:
## PR Checklist

Please check that your PR fulfills the following requirements:
- [x] The commit message follows [our guidelines](https://github.com/angular/components/blob/master/CONTRIBUTING.md#commit)
- [x] Tests for the changes have been added or weren't appropriate

## What is the current behavior?
Content on mobile devices overflows on certain pages in CDK.

Issue Number: 
#550 

## What is the new behavior?
Content does not overflow anymore.

## Other information

Before
Before state of CDK pages

After
After state of CDK pages

PR #657, opened 2019-10-17
fix api docs content from displaying after the left nav

andrewseguin:
Fixes issue introduced in #651


Splaktar:
It's not clear what the issue was, but based on the fix, it looks like there were some layout problems on mobile?

@donroyco FYI

Splaktar:
Oh, I see it...
Screen Shot 2019-10-17 at 23 00 12

Splaktar:
Also applying to `8.x` branch in PR https://github.com/angular/material.angular.io/pull/658.

donroyco:
Thanks @andrewseguin. How could I have overlooked this?

PR #660, opened 2019-10-18
add option for additional api docs for testing harnesses

andrewseguin:
Allows the API tab to include additional documentation files to be added to the page, making sure the table of contents also updates with them.

Primary objective is to allow the testing harness API docs to be appended to the component API page, e.g. button harness APIs will show up after the button APIs

### DIFF
--- a/src/app/pages/component-viewer/component-viewer.ts
+++ b/src/app/pages/component-viewer/component-viewer.ts
@@ -11,7 +11,7 @@ import {
 } from '@angular/core';
 import {MatTabsModule} from '@angular/material/tabs';
 import {ActivatedRoute, Params, Router, RouterModule} from '@angular/router';
-import {combineLatest, Observable, Subject} from 'rxjs';
+import {combineLatest, Observable, ReplaySubject, Subject} from 'rxjs';
 import {map, takeUntil} from 'rxjs/operators';
 import {DocViewerModule} from '../../shared/doc-viewer/doc-viewer-module';
 import {DocItem, DocumentationItems} from '../../shared/documentation-items/documentation-items';
@@ -26,7 +26,7 @@ import {ComponentPageTitle} from '../page-title/page-title';
   encapsulation: ViewEncapsulation.None,
 })
 export class ComponentViewer implements OnDestroy {
-  componentDocItem: DocItem;
+  componentDocItem = new ReplaySubject<DocItem>(1);
   sections: Set<string> = new Set(['overview', 'api']);
   private _destroyed = new Subject();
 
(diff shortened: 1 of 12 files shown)


src/app/pages/component-viewer/component-viewer.ts
```diff
@@ -63,47 +63,74 @@ export class ComponentViewer implements OnDestroy {
   }
 }
 
-@Component({
-  selector: 'component-overview',
-  templateUrl: './component-overview.html',
-  encapsulation: ViewEncapsulation.None,
-})
-export class ComponentOverview implements OnInit {
+export class ComponentBaseView implements OnInit, OnDestroy {
```
jelbourn: Add JsDoc?

src/app/pages/component-viewer/component-viewer.ts
```diff
@@ -63,47 +63,74 @@ export class ComponentViewer implements OnDestroy {
...
+      }
+    });
+  }
+
+  ngOnDestroy() {
+    this.destroyed.next();
   }
 
-  scrollToSelectedContentSection() {
+  onContentRendered(sectionName: string, docViewerContent: HTMLElement) {
```
jelbourn: Why change this name? I prefer naming functions based on what they do rather than when they're called

src/app/shared/table-of-contents/table-of-contents.ts
```diff
@@ -108,32 +129,13 @@ export class TableOfContents implements OnInit, AfterViewInit, OnDestroy {
...
-
-    return links;
-  }
-
   private onScroll(): void {
-    for (let i = 0; i < this.links.length; i++) {
-      this.links[i].active = this.isLinkActive(this.links[i], this.links[i + 1]);
+    const allLinks: Link[] = [];
+    this._linkSections.forEach(linkSection => {
+      allLinks.push(...linkSection.links);
```
jelbourn: Why reconstruct this array on every scroll?

src/app/pages/component-viewer/component-viewer.ts
andrewseguin: Changed since its scope of responsibility broadened, but since it's purpose is to make sure the Table of Contents is accurate, I can change to `updateTableOfContents`.

src/app/shared/table-of-contents/table-of-contents.ts
andrewseguin: Good catch, I refactored the function without looking at how often it's called. Changed so that the component keeps a cache on hand of the links it's displaying (new links appended when `addHeaders` is called, and reset in `resetHeaders`

src/app/shared/table-of-contents/table-of-contents.html
```diff
@@ -1,8 +1,8 @@
-<div *ngIf="links?.length" class="docs-toc-container">
-  <div class="docs-toc-heading">Contents</div>
+<div *ngFor="let linkSection of _linkSections" class="docs-toc-container">
+  <div class="docs-toc-heading">{{linkSection.name}}</div>
   <nav>
     <a [href]="_rootUrl + '#' + link.id"
```
Splaktar: Can we rename the harness `id`s to be `testing-classes` and `testing-interfaces`?

Currently, clicking on the links under "Testing" for "Classes" or "Interfaces" scrolls to the Component's Classes and Interfaces, not the Harness' Classes and Interfaces.

andrewseguin: Absolutely - I have a note to make sure we change the `dgeni` implementation so we do a better job of making the IDs unique in each doc (probably prefixed with the package name)

PR #673, opened 2019-11-07
build: update to new package name for examples package

devversion:
Updates the `material-examples` to `components-examples`. The new
example package includes the `ExampleModule` generated through a
genrule. This previously didn't work due to invalid metadata caused
by a bug in ngc.. but now since we build with Ivy.. this is no longer
a problem.


Splaktar:
@devversion which Components issue # does this change map to?

I'm going to merge this as it's blocking people from building this project on `master` with the latest docs-content.

PR #685, opened 2019-12-03
build: update to rc.4. re-run 8->9 migrations.

Splaktar:
- update to 4.x GA of MDC Web to match components repo
- remove hammerjs via migration
- remove hammerjs from StackBlitz template
- update dev dependencies


Splaktar:
@devversion I know that we don't fix all tests for HammerJS, but it seems like we should maybe remove the import from `src/test.ts` if we remove the `hammerjs` dependency from the repo?

devversion:
@Splaktar This pattern was never recommended (unless I missed it somewhere), nor do we set it up through our ng-add schematics.. so I figured we don't want to handle that. 

We could totally handle such imports, but the migration is already quite complex. Also it's probably better if the import causes the tests to fail, than just removing it (which could cause confusing error messages)

PR #687, opened 2019-12-05
fix: header not working correctly

### LINKED ISSUE #686: On some pages the header is not taking full width
On some pages of the material.angular.io, the header is not taking full width on some screens. I've attached a screenshot for reference.

Datepicker   Angular Material (2)

ajitsinghkaler:
Fixes #686

The header is not working correctly when content takes more width than the viewport. Check the attached screenshot. Now make the body scroll if it takes more with than the screen.

Datepicker   Angular Material (2)

### DIFF
--- a/src/app/pages/component-sidenav/component-sidenav.scss
+++ b/src/app/pages/component-sidenav/component-sidenav.scss
@@ -115,6 +115,12 @@ app-component-sidenav {
 .docs-component-sidenav-body-content {
   display: flex;
   flex: 1 1 auto;
+  overflow-x: auto;
+  overflow: -moz-scrollbars-none;
+}
+
+.docs-component-sidenav-body-content::-webkit-scrollbar {
+  display: none;
 }
 
 @media (max-width: $small-breakpoint-width) {


Splaktar:
Can you please take a look at the issue and fix in https://github.com/angular/material.angular.io/pull/657 to make sure that we aren't re-introducing the same issue on desktop here?

ajitsinghkaler:
After checking I found out that this does not lead to the same issues you can also have a check

ajitsinghkaler:
@crisbeto can you please review this.

crisbeto:
On what browser does the original issue happen on? I tried against Chrome, Firefox and IE, but I couldn't reproduce it after resizing the window down so the example doesn't fit.

ajitsinghkaler:
@crisbeto it was happen ing on all browsers. I checked on Chrome and Mozilla. I'm also attaching a video link for your reference.

Steps to reproduce open: https://material.angular.io/components/datepicker/overview
Reduce your screen width to 737px in responsive mode(Actually happens in other screens but apparent on this screen).
try to scroll left-right. You will see the error.

https://drive.google.com/file/d/1W49_QHKFNpWUqgmyWqNLv1mFSh6SEejt/view?usp=drivesdk

ajitsinghkaler:
@crisbeto please tell me if you are not able to recreate it.

crisbeto:
ok, I see what you mean now. I'm not sure that this the way to go though. If you scroll all the way to the bottom you can see that we now have an extra scrollbar:

ajitsinghkaler:
@crisbeto you have any other suggestions that is what I intended to do.

crisbeto:
Ideally it should stretch out to the edge of the screen and let the page have the horizontal overflow. I'm having a hard time to behave that way though.

ajitsinghkaler:
@crisbeto Should we hide the scrollbars via css does that seem like a valid solution

PR #688, opened 2019-12-13
build: enable strictTemplates, strictBindCallApply, and strictFunctionTypes

Splaktar:
- enable Ivy's strict Template Type Checking
- fix related build errors
- complete `this._destroyed` in `ngOnDestroy()`
- exclude `dist/` from linting

`strictPropertyInitialization` is a bit more problematic, so it will be handled in a separate PR.

### DIFF
--- a/src/app/shared/table-of-contents/table-of-contents.ts
+++ b/src/app/shared/table-of-contents/table-of-contents.ts
@@ -105,9 +105,9 @@ export class TableOfContents implements OnInit, AfterViewInit, OnDestroy {
   }
 
   addHeaders(sectionName: string, docViewerContent: HTMLElement) {
-    const headers = docViewerContent.querySelectorAll('h3, h4');
+    const headers = Array.from<HTMLHeadingElement>(docViewerContent.querySelectorAll('h3, h4'));
     const links: Link[] = [];
-    headers.forEach((header: HTMLElement) => {
+    headers.forEach((header) => {
       // remove the 'link' icon name from the inner text
       const name = header.innerText.trim().replace(/^link/, '');
       const {top} = header.getBoundingClientRect();
--- a/src/app/shared/theme-picker/theme-picker.ts
+++ b/src/app/shared/theme-picker/theme-picker.ts
@@ -1,13 +1,13 @@
 import {
-  Component,
-  ViewEncapsulation,
   ChangeDetectionStrategy,
+  Component,
   NgModule,
-  OnInit,
   OnDestroy,
+  OnInit,
+  ViewEncapsulation,
 } from '@angular/core';
 import {StyleManager} from '../style-manager';
-import {ThemeStorage, DocsSiteTheme} from './theme-storage/theme-storage';
+import {DocsSiteTheme, ThemeStorage} from './theme-storage/theme-storage';
 import {MatButtonModule} from '@angular/material/button';
 import {MatGridListModule} from '@angular/material/grid-list';
 import {MatIconModule} from '@angular/material/icon';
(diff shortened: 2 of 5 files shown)


src/app/shared/table-of-contents/table-of-contents.ts
```diff
@@ -107,9 +107,9 @@ export class TableOfContents implements OnInit, AfterViewInit, OnDestroy {
   addHeaders(sectionName: string, docViewerContent: HTMLElement) {
     const headers = docViewerContent.querySelectorAll('h3, h4');
     const links: Link[] = [];
-    headers.forEach((header: HTMLElement) => {
+    headers.forEach((header: Element) => {
```
jelbourn: I would instead do
```ts
const headers = Array.from(docViewerContent.querySelectorAll('h3, h4')) as HTMLElement[];
```
and then the type of the types should be inferred

jelbourn:
I'm also not really concerned w/ turning on `strictPropertyInitialization`, which doesn't play super well w/ Angular in general

src/app/shared/table-of-contents/table-of-contents.ts
```diff
@@ -105,9 +105,9 @@ export class TableOfContents implements OnInit, AfterViewInit, OnDestroy {
   }
 
   addHeaders(sectionName: string, docViewerContent: HTMLElement) {
-    const headers = docViewerContent.querySelectorAll('h3, h4');
+    const headers = Array.from(docViewerContent.querySelectorAll('h3, h4')) as HTMLHeadingElement[];
```
crisbeto: It's pretty much the same, but I think you can do this which is a little shorter `Array.from<HTMLHeadingElement>(docViewerContent.querySelectorAll('h3, h4'))`.

src/app/shared/theme-picker/theme-picker.ts
```diff
@@ -71,8 +70,12 @@ export class ThemePicker implements OnInit, OnDestroy {
 
   ngOnInit() {
     this._queryParamSubscription = this._activatedRoute.queryParamMap
-      .pipe(map((params: ParamMap) => params.get('theme')), filter(Boolean))
-      .subscribe((themeName: string) => this.installTheme(themeName));
+      .pipe(map((params: ParamMap) => params.get('theme')))
```
crisbeto: I think the previous approach was valid, you just needed to do `themeName!` since it's guaranteed to be defined.

Splaktar:
```ts
    this._queryParamSubscription = this._activatedRoute.queryParamMap
    .pipe(map((params: ParamMap) => params.get('theme')), filter(Boolean))
    .subscribe((themeName: string) => this.installTheme(themeName!));
```
Results in:
```ts
ERROR in src/app/shared/theme-picker/theme-picker.ts:74:16 - error TS2769: No overload matches this call.
  Overload 1 of 5, '(observer?: NextObserver<unknown> | ErrorObserver<unknown> | CompletionObserver<unknown> | undefined): Subscription', gave the following error.
    Argument of type '(themeName: string) => void' is not assignable to parameter of type 'NextObserver<unknown> | ErrorObserver<unknown> | CompletionObserver<unknown> | undefined'.
      Property 'complete' is missing in type '(themeName: string) => void' but required in type 'CompletionObserver<unknown>'.
  Overload 2 of 5, '(next?: ((value: unknown) => void) | undefined, error?: ((error: any) => void) | undefined, complete?: (() => void) | undefined): Subscription', gave the following error.
    Argument of type '(themeName: string) => void' is not assignable to parameter of type '(value: unknown) => void'.
      Types of parameters 'themeName' and 'value' are incompatible.
        Type 'unknown' is not assignable to type 'string'.

74     .subscribe((themeName: string) => this.installTheme(themeName!));
                  ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

  node_modules/rxjs/internal/types.d.ts:64:5
    64     complete: () => void;
           ~~~~~~~~
    'complete' is declared here.
```
This compiles, but is not great...
```ts
.subscribe((themeName: any) => this.installTheme(themeName));
```

PR #689, opened 2019-12-14
build: enable strictPropertyInitialization

Splaktar:
- fix related build errors
- complete `this._destroyed` in `ngOnDestroy()`
- replace use of deprecated `DomPortalHost` with `DomPortalOutlet`

Builds on PR https://github.com/angular/material.angular.io/pull/688

### DIFF
--- a/src/app/shared/doc-viewer/doc-viewer.ts
+++ b/src/app/shared/doc-viewer/doc-viewer.ts
@@ -26,9 +26,9 @@ import {HeaderLink} from './header-link';
 })
 export class DocViewer implements OnDestroy {
   private _portalHosts: DomPortalOutlet[] = [];
-  private _documentFetchSubscription: Subscription;
+  private _documentFetchSubscription: Subscription = new Subscription();
 
-  @Input() name: string;
+  @Input() name: string | undefined;
 
   /** The URL of the document to display. */
   @Input()
--- a/src/app/pages/component-category-list/component-category-list.ts
+++ b/src/app/pages/component-category-list/component-category-list.ts
@@ -1,13 +1,18 @@
+import {CommonModule} from '@angular/common';
 import {Component, NgModule, OnDestroy, OnInit} from '@angular/core';
 import {MatCardModule} from '@angular/material/card';
-import {CommonModule} from '@angular/common';
 import {ActivatedRoute, Params, RouterModule} from '@angular/router';
-import {DocumentationItems, SECTIONS} from '../../shared/documentation-items/documentation-items';
-import {ComponentPageTitle} from '../page-title/page-title';
-import {SvgViewerModule} from '../../shared/svg-viewer/svg-viewer';
-import {Observable, combineLatest, Subscription} from 'rxjs';
-import {NavigationFocusModule} from '../../shared/navigation-focus/navigation-focus';
+import {combineLatest, Observable, Subscription} from 'rxjs';
 
+import {
+  DocumentationItems,
+  SECTIONS
+} from '../../shared/documentation-items/documentation-items';
+import {
+  NavigationFocusModule
+} from '../../shared/navigation-focus/navigation-focus';
+import {SvgViewerModule} from '../../shared/svg-viewer/svg-viewer';
+import {ComponentPageTitle} from '../page-title/page-title';
 
 @Component({
   selector: 'app-component-category-list',
@@ -15,9 +20,9 @@ import {NavigationFocusModule} from '../../shared/navigation-focus/navigation-fo
   styleUrls: ['./component-category-list.scss']
 })
 export class ComponentCategoryList implements OnInit, OnDestroy {
-  params: Observable<Params>;
-  routeParamSubscription: Subscription;
-  _categoryListSummary: string;
+  params: Observable<Params> | undefined;
+  routeParamSubscription: Subscription = new Subscription();
+  _categoryListSummary: string | undefined;
 
   constructor(public docItems: DocumentationItems,
               public _componentPageTitle: ComponentPageTitle,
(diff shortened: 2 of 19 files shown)


Splaktar:
@crisbeto can you please take a look at this [Circle CI](https://circleci.com/gh/angular/material.angular.io/441) error? It seems to be related to the [Portal Changes in 9.0.0](https://github.com/angular/components/pull/16708).

Should I file a bug here or is there some misconfiguration in this PR?

src/app/shared/doc-viewer/doc-viewer.ts
```diff
@@ -19,16 +19,16 @@ import {Subscription} from 'rxjs';
 import {take} from 'rxjs/operators';
 import {ExampleViewer} from '../example-viewer/example-viewer';
 import {HeaderLink} from './header-link';
+import {DomPortalOutlet} from '@angular/cdk/portal/dom-portal-outlet';
```
devversion: As mentioned on Slack. This import seems the culprit. Any reason why you use a deep import?

Splaktar: That's what WebStorm defaulted to when auto importing 😞

Splaktar:
I just found that this branch/PR were passing on CircleCI, but failing to build locally. It appears that `ng test` doesn't expose Ivy's Template Type Checking errors. However, Template Type Checking does break the build. So I've added a build task to CI to catch this issue.

Additionally, it also runs `build:content` to ensure that the docs-content can be synced as we've had a couple issues this year where that was broken and we weren't aware of it immediately.

Splaktar:
OK, the issues that I'm hitting were introduced in Angular `9.0.0-rc.5` as part of PR https://github.com/angular/angular/pull/33997 (specifically https://github.com/angular/angular/pull/33997/commits/66e8ed4ca8c7bd140e7dd3de0493e21382d2d755) which was fixing https://github.com/angular/angular/issues/31556.

I'm investigating how to solve them, but I don't feel like my current solution is solid, so I'm continuing to investigate.

Splaktar:
@crisbeto or @JoostK do you have any guidance here?

### Template
```html
<ng-container *ngIf="componentViewer.componentDocItem | async; let docItem">
  <span class="cdk-visually-hidden" tabindex="-1" #initialFocusTarget>
    Examples for {{docItem.id}}
  </span>
  <example-viewer *ngFor="let example of docItem.examples"
                  [example]="example"></example-viewer>
</ng-container>
```
There are errors for
```
src/app/pages/component-viewer/component-examples.html:3:20 - error TS2531: Object is possibly 'null'.

3     Examples for {{docItem.id}}
                     ~~~~~~~~~~

  src/app/pages/component-viewer/component-viewer.ts:132:16
    132   templateUrl: './component-examples.html',
                       ~~~~~~~~~~~~~~~~~~~~~~~~~~~
    Error occurs in the template of component ComponentExamples.
src/app/pages/component-viewer/component-examples.html:5:34 - error TS2531: Object is possibly 'null'.

5   <example-viewer *ngFor="let example of docItem.examples" [example]="example"></example-viewer>
                                   ~~~~~~~~~~~~~~~~

  src/app/pages/component-viewer/component-viewer.ts:132:16
    132   templateUrl: './component-examples.html',
                       ~~~~~~~~~~~~~~~~~~~~~~~~~~~
    Error occurs in the template of component ComponentExamples.
```

# Solutions?

It's complaining about `docItem.id` and `docItem.examples` where `docItem` might be `null` in these cases. But it seems to be ignoring the fact that this template won't be rendered if `docItem` is `null`.

The tests in PR https://github.com/angular/angular/pull/33997 don't cover the use case where the Async Pipe is used.

How do you recommend developers solve this very common use case?

JoostK:
@Splaktar ah, this is interesting. Using `as` the way we tested is reflects the `NgIfContext.ngIf` property, for which the template type checker has applied the appropriate context guard expression due to the presence of `NgIf.ngContextGuard_ngIf`.

When used with a let assignment to `$implicit`, however, the context guard for `ngIf` does not narrow reads of `$implicit`.

I'll have a look to see if there's something that can be done here.

Splaktar:
@JoostK I tried the following and got the same errors:
```html
<ng-container *ngIf="componentViewer.componentDocItem | async as docItem">
```

Splaktar:
@JoostK here is the [workaround](https://github.com/angular/material.angular.io/commit/f813cc6d0c12bc52dab5237922403fcde4108e72) that I had to apply to get our repo building again after the breaking change in Angular `9.0.0-rc.5`. Do you have an open GitHub issue to track this or should I open one?

JoostK:
@Splaktar there isn't any issue at the moment, so please open one.

In the meantime I am working on a PR to resolve this issue, but I'm not fond of the approach quite yet.

The main reason as to why template variables are not properly narrowed is because [the "template guard" for the `ngIf` input](https://github.com/angular/angular/blob/cca26166376d4920f5905b168e70ea2e8d70da77/packages/common/src/directives/ng_if.ts#L227) is only regarding the input binding expression itself. When using template variables, however, the value is taken from `NgIfContext` which is _not_ narrowed as it's currently not possible to insert template guards for context variables. The PR I'm working on changes [the "template context guard" of `NgIf`](https://github.com/angular/angular/blob/cca26166376d4920f5905b168e70ea2e8d70da77/packages/common/src/directives/ng_if.ts#L235) to let the context type be inferred to `NgIfContext<NonNullable<T>>` so that its properties `$implicit` and `ngIf` would no longer include the nullable types.

Splaktar:
Opened https://github.com/angular/angular/issues/34572 to track the issue with ngIf + async pipe type narrowing.

Splaktar:
PR https://github.com/angular/material.angular.io/pull/926 reverts the workarounds, mentioned above, to get the builds working with the async pipe type narrowing bug.

Splaktar:
@jelbourn this is the PR I mentioned in the meeting and you said that you would take a quick look at. This would get us closer to being aligned with the Angular CLI's strict mode.

andrewseguin (approved):
LGTM

PR #691, opened 2019-12-19
build: update Angular to 9.0.0-rc.7 

Splaktar:
- update firebase-tools, @types/node, codelyzer

ci: ensure that the project builds successfully. update NodeJS
- update NodeJS v10 to v12
- test that both installing the docs-content and building pass
  - Template Type Checking errors aren't emitted for `ng test`
  - so it's possible to have lint and test pass when the project doesn't build
  - we've also had issues this year where build:content was broken and we didn't notice


.circleci/config.yml
```diff
@@ -12,26 +12,46 @@ var_4: &save_cache
   save_cache:
     key: *cache_key
     paths:
-    - "node_modules/"
+    - "node_modules"
```
devversion: I personally prefer the trailing slash as it makes it clear that this is a directory and _not_ a file. Not a big deal though 😄

Splaktar: 😀 I actually made that change to align with what's in `angular/components` 😉

devversion: Haha, you got me there 😄 No for real, next time I'll change things there I'll make the change there!

Splaktar: Sounds good! 😆

Issue #671, opened 2019-10-31
Assorted accessibility issues


jelbourn:
These issues were forwarded to me from a user that relies on assistive technology and was trying to use the site

### 1. The color contrast of everything in the left menu is insufficient.
That is generating 50+ Axe violations.

* The gray text has a contrast ratio of 3.88:1 and needs to be 4.5.1
 https://webaim.org/resources/contrastchecker/?fcolor=797979&bcolor=F2F2F2

* The blue text is dark enough
https://webaim.org/resources/contrastchecker/?fcolor=3F51B5&bcolor=F2F2F2

### 2. In the left menu, the currently selected item is indicated
only by a color change, you can never use color alone to communicate
information.

* For sighted users add an underline, arrow or dot for the selected
item (it's fine to keep the color too).
* For screen reader users, add aria-current="true" or
aria-current="page", either is fine.

### 3. Missing "accessible name" on action buttons.
       The “view source” buttons have no screen reader label

* The “Edit in Stack Blitz” buttons have no screen reader label.
Use aria-label to add these, e.g. <button aria-label="view source"....
Ideally these labels should be unique, so if you have multiple
examples on the same page the aria-label should be "view example 1
source code" but that's best practice.

### 4. The “Select a theme” button has a tooltip, but it does not have an
accessible name – the tooltip info is the name, but has been put into
aria-describedby which maps it as a description, not a label. Just
change "aria-describedby" to "aria-labelledby" (yes, 2 ls) and the
issue is fixed.

### 5. (headings/landmarks)
I) There is no heading or landmark that marks the start of the unique
content on an example page, which makes it hard for a screen reader
user to locate it.
There is a visually hidden <span> with the title, but it needs
role="heading" and aria-level="2" to be marked as an h2 heading.
E.g. (from: https://material.angular.io/components/form-field/overview)
```
<component-overview class="ng-star-inserted"
<!-- added role="heading" and aria-level="2" to make this an h2 heading). -->

><span tabindex="-1" class="cdk-visually-hidden" role="heading" aria-level="2"> Overview for form-field
</span>
```

II) Most headings have a link that is announced before the heading
content and makes them extremely annoying to navigate with a screen
reader:
```
<h3 id="form-field-appearance-variants" class="docs-header-link">

          <span header-link="form-field-appearance-variants">
<header-link class="ng-star-inserted">
<!-- This link needs to be moved outside the heading, preferably after
though before is acceptable, it just can't be part of the heading tag, ,
[...]

PR #697, opened 2019-12-28
docs-infra: address numerous a11y issues

Splaktar:
Please see commits for descriptions and related issues (https://github.com/angular/material.angular.io/issues/671 captures most, but not all of the issues).

## More Accessible Theme Picker
The current theme picker is a `mat-grid-list` nested in a `mat-menu` which isn't ideal.
The new theme picker is a `mat-menu` with `button` elements.

Screen Shot 2019-12-28 at 19 34 31
Screen Shot 2019-12-28 at 19 34 13

## TODO
- [x] [This list](https://github.com/angular/material.angular.io/issues/671#issuecomment-569379349)
- [x] Add some visual color hints to theme picker


src/app/shared/theme-picker/theme-picker.ts
```diff
@@ -82,7 +90,7 @@ export class ThemePicker implements OnInit, OnDestroy {
     this._queryParamSubscription.unsubscribe();
   }
 
-  installTheme(themeName: string) {
+  onThemeSelection(themeName: string) {
     const theme = this.themes.find(currentTheme => currentTheme.name === themeName);
 
     if (!theme) {
```
Splaktar: `MatGridListModule` probably doesn't need to be imported into this module anymore.

src/app/pages/component-viewer/component-viewer.html
```diff
@@ -1,5 +1,5 @@
 <div class="docs-component-viewer">
-  <nav mat-tab-nav-bar class="docs-component-viewer-tabbed-content">
+  <nav mat-tab-nav-bar class="docs-component-viewer-tabbed-content" aria-label="Section">
```
jelbourn: I'm not sure that "Section" is sufficient on its own

src/app/shared/stack-blitz/stack-blitz-button.html
```diff
@@ -1,7 +1,6 @@
-<div [matTooltip]="isDisabled ? 'Building StackBlitz example...' : 'Edit in StackBlitz'">
-  <button mat-icon-button type="button"
-          (click)="openStackBlitz()"
-          [disabled]="isDisabled">
-    <mat-icon>open_in_new</mat-icon>
-  </button>
-</div>
+<button mat-icon-button type="button" (click)="openStackBlitz()" [disabled]="isDisabled"
+        [attr.aria-label]="isDisabled ? 'Building StackBlitz example...' : 'Edit in StackBlitz'"
```
jelbourn: Maybe "Edit this example in StackBlitz"

src/app/shared/table-of-contents/table-of-contents.html
```diff
@@ -1,6 +1,6 @@
 <div *ngFor="let linkSection of _linkSections" class="docs-toc-container">
   <div class="docs-toc-heading">{{linkSection.name}}</div>
-  <nav>
+  <nav aria-label="Section Table of Contents">
```
jelbourn: Instead of "Section", could we use the name of the item being documented at the time? Maybe something like "Button overview table of contents"

src/app/shared/theme-picker/theme-picker.html
```diff
@@ -1,18 +1,15 @@
...
-      <div mat-menu-item (click)="installTheme(theme.name)">
-        <div class="docs-theme-picker-swatch">
-          <mat-icon class="docs-theme-chosen-icon" *ngIf="currentTheme === theme">check_circle</mat-icon>
-          <div class="docs-theme-picker-primary" [style.background]="theme.primary"></div>
-        </div>
-      </div>
-    </mat-grid-tile>
-  </mat-grid-list>
+<mat-menu #themeMenu="matMenu" xPosition="before" class="theme-picker-menu">
+  <button mat-menu-item *ngFor="let theme of themes" (click)="onThemeSelection(theme.name)">
```
jelbourn: This button also needs an `aria-label`

src/app/shared/theme-picker/theme-picker.scss
```diff
@@ -1,58 +1,62 @@
...
-    margin: ($theme-picker-grid-cell-size - $theme-picker-swatch-size) / 2;
-    border-radius: 50%;
-    overflow: hidden;
-
-    .docs-theme-chosen-icon {
-      color: white;
-      position: absolute;
-      left: 50%; top: 50%;
-      transform: translate(-50%, -50%);
+.theme-picker-menu {
```
jelbourn: Prefix with `docs-`

src/app/shared/theme-picker/theme-picker.spec.ts
```diff
@@ -15,7 +15,7 @@ describe('ThemePicker', () => {
     const component = fixture.componentInstance;
     const name = 'pink-bluegrey';
     spyOn(component.styleManager, 'setStyle');
-    component.installTheme(name);
+    component.onThemeSelection(name);
```
jelbourn: I prefer the previous name; my philosophy is that method names should describe what they do, not how they're called

src/assets/img/theme-demo-icon.svg
```diff
@@ -0,0 +1,28 @@
...
+        <use xlink:href="#button" opacity="1" fill="#ff4081" fill-opacity="1"
+             class="theme-icon-button"></use>
+      </g>
+      <g>
+        <use xlink:href="#toolbar" opacity="1" fill="#3f51b5" fill-opacity="1"
+             class="theme-icon-toolbar"></use>
+      </g>
+    </g>
+  </g>
+</svg>
```
jelbourn: Run this through `svgo`?

src/app/shared/table-of-contents/table-of-contents.html
Splaktar: We can use `linkSection.name` in the label here, but that would give the following
- "Overview content Table of Contents" for a component's overview TOC
- "Checkbox Table of Contents" for a component's API TOC

I suppose that this is better than just "Section". If we want to add something like "Checkbox overview", we would need to add another Input and pass that data in as we don't currently have that available in the TOC component.

src/app/shared/theme-picker/theme-picker.html
Splaktar: The screen reader announces the content of the button which is in the `<span>` on line 12. I.e. "Deep Purple slash Amber, menu item 1 of 4". That seems sufficient?

src/assets/img/theme-demo-icon.svg
Splaktar: svgo removes `viewBox="0 0 80 80"` which breaks the icons, but adding it back seems to have fixed it. The rest of it's changes seem like valid minifications w/o side effects.

src/app/shared/table-of-contents/table-of-contents.html
jelbourn: I think that is better, though adding another input seems pretty straightforward

src/app/shared/theme-picker/theme-picker.html
jelbourn: I missed that span

src/app/shared/theme-picker/theme-picker.ts
```diff
@@ -9,62 +9,69 @@ import {
...
 export class ThemePicker implements OnInit, OnDestroy {
   private _queryParamSubscription = Subscription.EMPTY;
   currentTheme: DocsSiteTheme;
 
+  // The below colors need to align with the themes defined in theme-picker.scss
   themes: DocsSiteTheme[] = [
     {
       primary: '#673AB7',
       accent: '#FFC107',
+      displayName: 'Deep Purple / Amber',
```
jelbourn: We should probably use either "and" or "&" rather than "/"

src/app/shared/theme-picker/theme-picker.scss
```diff
@@ -1,58 +1,62 @@
...
+      margin-left: 8px;
+      float: right;
+
+      svg {
+        vertical-align: middle;
+      }
+
+      // The below colors need to align with the themes defined in ThemePicker
+      &.deeppurple-amber svg {
+        .theme-icon-background {
```
jelbourn: These classes still need a `docs-` prefix

Splaktar:
Addressed feedback and also added a LiveAnnouncer message when a theme is selected from the list.

PR #699, opened 2020-01-09
build: update to Angular and CLI v9 rc.8 and Components v9 rc.7

Splaktar:


Splaktar:
Rebased to pick up PR https://github.com/angular/material.angular.io/pull/700 which updated the same Components dependencies to rc.7.

PR #705, opened 2020-01-18
fix: numerous responsive issues on screens from 360px to 960px

### LINKED ISSUE #550: Mobile layout issues with some CDK pages
This is not a great mobile experience. CDK docs only. Material docs are fine. 

No page should be horizontally scrollable and text/examples should wrap properly on mobile.

Affects
- [ ] drag and drop
- [x] bidi
- [ ] observers
- [ ] accessibility
- [ ] check for others

screenshot_20181020-044549

### LINKED ISSUE #686: On some pages the header is not taking full width

Splaktar:
#### fix(component-sidenav): responsive issues w/ screens 720px to 960px width

- close over sidenav on nav item selection

Fixes #686. Closes #687.

#### fix(docs-api): tables overflow screen width on mobile

- fix docs-api layouts to work down to 360px width
  - fixes Observers, Accessibility, Drag and Drop, Platform, and Overlay layouts
- fix DeprecatedconnectedTo to Deprecated connectedTo display issue
- update footer copyright
- fix exception trying to unsubscribe to undefined routeParamSubscription

Fixes #550

There is still an issue with the Datepicker Overview page, but that's an issue with https://github.com/angular/components/blame/dd4295608a9763518cd40b43b94059cc9c606fb4/src/material/datepicker/datepicker.md#L214-L218 rather than the docs site. That markdown table needs to be replaced with something that will work more responsively. This is fixed in PR https://github.com/angular/components/pull/18210.

### Before
Screen Shot 2020-01-18 at 07 39 18

### After
Screen Shot 2020-01-18 at 07 39 10


src/app/pages/component-sidenav/component-nav.html
```diff
@@ -10,12 +10,12 @@
       </button>
       <mat-nav-list dense [@bodyExpansion]="panel.expanded ? 'expanded' : 'collapsed'"
                     id="panel-{{category.id}}" [attr.aria-label]="category.name">
-        <a mat-list-item *ngFor="let component of category.items"
-           [routerLink]="'/' + (params | async)?.section+ '/' + component.id"
-           routerLinkActive="docs-component-viewer-sidenav-item-selected"
-           [attr.aria-current]="currentItemId === component.id ? 'page': 'false'">
+        <button mat-list-item *ngFor="let component of category.items"
```
jelbourn: These should remain `<a>` elements:
* The `<a>` element signals to screen readers that this is a link and says where it goes
* You can't ctrl + click or shift + click a button
* You can't right click and copy the URL location on a button

src/app/pages/component-sidenav/component-sidenav.service.ts
```diff
@@ -0,0 +1,13 @@
+import {Injectable} from '@angular/core';
+import {MatSidenav} from '@angular/material/sidenav';
+
+@Injectable({providedIn : 'root'})
+export class ComponentSidenavService {
```
jelbourn: What is the responsibility of this class? 
* Should have a class description comment
* Should not use "Service" in the name since it doesn't communicate anything about what it does; I always want us to use class names that describe the responsibility
(e.g., Angular Material doesn't have "RadioService", but instead "UniqueSelectionDispatcher")

src/app/pages/component-sidenav/component-sidenav.service.ts
```diff
@@ -0,0 +1,13 @@
...
+  private _sidenav: MatSidenav;
+
+  constructor() {}
```
jelbourn: Can delete empty constructor

src/app/pages/component-sidenav/component-sidenav.service.ts
```diff
@@ -0,0 +1,13 @@
...
+
+  get sidenav(): MatSidenav { return this._sidenav; }
+
+  set sidenav(value: MatSidenav) { this._sidenav = value; }
```
jelbourn: We try to only use getters/setters for `@Input` properties. This one looks like it could just be a regular property.

src/app/pages/component-sidenav/component-sidenav.ts
```diff
@@ -37,9 +37,12 @@ import {HttpClientModule} from '@angular/common/http';
 import {StackBlitzButtonModule} from '../../shared/stack-blitz';
 import {SvgViewerModule} from '../../shared/svg-viewer/svg-viewer';
 import {ExampleModule} from '@angular/components-examples';
+import {ComponentSidenavService} from './component-sidenav.service';
+import {MatDrawerToggleResult} from '@angular/material/sidenav/drawer';
 import {MatListModule} from '@angular/material/list';
 
-const SMALL_WIDTH_BREAKPOINT = 720;
+const EXTRA_SMALL_WIDTH_BREAKPOINT = 720;
```
jelbourn: Add comment that explains how you arrived at these values?
(at least for the new one, not sure where the original 720 came from)

src/app/pages/component-sidenav/component-sidenav.service.ts
Splaktar: This service allows for the `ComponentSidenav` and the `ComponentNav` to share access to the `MatSidenav` instance when it is open (it needs to be re-set each time the "over" sidenav is opened). This allows the `ComponentNav` to close the "over" sidenav on mobile when a nav item selection is made.

I guess something like `ComponentSidenavCoordinator` would desribe its responsibility?

src/app/pages/component-sidenav/component-nav.html
Splaktar: I agree with all of your bullet points and had those in mind when making this change.

However the mobile UX with links in the sidenav is poor because clicking a link leaves the sidenav open and the backdrop covering the page. This requires an extra tap + the knowledge of where to tap in order to see the new content. It's also possible that the user won't realize that navigation has occurred since there is no ripple, only a brief rectangular flash, the nav item text color changes, but the content is almost completely obscured behind the "over" sidenav.

With the changes to close the "over" sidenav on selection, the animation was extremely janky and ugly. Changing them to `button`s and waiting to navigate until the sidenav close Promise resolved in `selectNavItem()` was the only way that I could find to remove the janky animation.

Keeping the `<a>` tag and using `(click)="selectNavItem(component.id)"` to dismiss the "over" sidenav was janky as the navigation always started eagerly and larger page loads (like select) would jank the sidenav closing animation.

I guess that I could investigate waiting for a `NavigationEnd` event from the Router before I closed the sidenav. I'll take a look.

src/app/pages/component-sidenav/component-sidenav.service.ts
jelbourn: Could the `ComponentSidenav` grab the `MatSidenav` with a static `ViewChild` so that the `ComponentNav` could inject its parent and access it that way without adding another class?

src/app/pages/component-sidenav/component-nav.html
jelbourn: Yeah, that's where my thoughts go as well

src/app/pages/component-sidenav/component-sidenav.service.ts
Splaktar: I'll give it a shot.

src/app/pages/component-sidenav/component-nav.html
Splaktar: It looks like the `NavigationEnd` router event is fired when the router is done with that lifecycle step, but not when the browser is done rendering the content. Thus it still janks on large page renders. 

It's unfortunate that we lose the ability to close the sidenav smoothly when using `routerLink` versus button `(click)` events, which give us full control of the order of operations, so that we can ensure that the sidenav closes first via `sidenav.close().then(() => doNav());`

I don't have any other ideas at this time.  For now, I think that it's better to close the sidenav on navigation, even if there is some jank, rather than forcing the mobile user to tap the backdrop to get to the content.

src/app/pages/component-sidenav/component-sidenav.service.ts
Splaktar: Removed the whole service since it is not needed when using anchors and `routerLink` to navigate.

src/app/pages/component-sidenav/component-nav.html
Splaktar: Hrm. I did a production build and did some testing on Firebase Hosting (https://angularmaterial.dev/) and I'm no longer seeing any jank when using `NavigationEnd`. Perhaps it was just a side-effect of devMode.

PR #712, opened 2020-02-06
fix(stack-blitz): default to form-field fill appearance

Splaktar:
Can be tested via https://angularmaterial.dev/components/datepicker/examples

Relates to https://github.com/angular/components/issues/14792 and https://github.com/angular/components/pull/18252.

This should go into `master` ASAP for the v9 final release.

### DIFF
--- a/src/assets/stack-blitz/src/main.ts
+++ b/src/assets/stack-blitz/src/main.ts
@@ -8,9 +8,13 @@ import {BrowserModule} from '@angular/platform-browser';
 import {platformBrowserDynamic} from '@angular/platform-browser-dynamic';
 import {BrowserAnimationsModule} from '@angular/platform-browser/animations';
 import {DemoMaterialModule} from './app/material-module';
+import {MAT_FORM_FIELD_DEFAULT_OPTIONS} from '@angular/material/form-field';
 
 import {MaterialDocsExample} from './app/material-docs-example';
 
+// Default MatFormField appearance to 'fill' as that is the new recommended approach and the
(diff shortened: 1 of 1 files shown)


src/assets/stack-blitz/src/main.ts
```diff
@@ -24,7 +25,9 @@ import {MaterialDocsExample} from './app/material-docs-example';
   entryComponents: [MaterialDocsExample],
   declarations: [MaterialDocsExample],
   bootstrap: [MaterialDocsExample],
-  providers: []
+  providers: [
+    { provide: MAT_FORM_FIELD_DEFAULT_OPTIONS, useValue: { appearance: 'fill' } },
```
devversion: Can you please add a comment on why we do this?

Splaktar: Added, PTAL.

README.md
```diff
@@ -3,11 +3,11 @@
 This is the repository for the [Angular Components documentation site](https://material.angular.io/).
 
 Versions of this site are also available for
-- [Beta](https://next.material.angular.io/)
 - [v5](https://v5.material.angular.io/)
 - [v6](https://v6.material.angular.io/)
 - [v7](https://v7.material.angular.io/)
 - [v8](https://material.angular.io/)
+- [v9-rc](https://next.material.angular.io/)
```
jelbourn: Can you revert this? I'm about to update it for actual v9

Splaktar: Done

PR #714, opened 2020-02-06
Update everything to v9

jelbourn:


Splaktar:
Stackblitz template needs to be updated to `^9.0.0`.

jelbourn:
@Splaktar good catch, done

PR #716, opened 2020-02-07
fix: cdk test harness API document not found

devversion:
The CDK test harness documentation API has `test-harnesses` as id. This
doesn't match any generated API document since the API doc for
`@angular/cdk/testing` is generated with the `cdk-testing` id.

We should add support for custom API doc ids that differ from the
actual doc entry id in order to fix this issue.

Fixes angular/components#18426

### DIFF
--- a/src/app/pages/component-viewer/component-api.html
+++ b/src/app/pages/component-viewer/component-api.html
@@ -3,8 +3,7 @@
     API for {{docItem?.id}}
   </span>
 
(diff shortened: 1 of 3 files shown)


devversion:
This is a way we could solve this issue. I think it's reasonable since we obviously have doc items that don't necessarily match an entry-point. i.e. `test-harnesses` is not purely `@angular/cdk/testing`.

PR #724, opened 2020-02-20
build: use better resource class for building in circleci

devversion:
We generate a lot of chunks with our build. To improve stability and to reduce the amount it takes to run, we use a higher resource class with better VM specs.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -40,6 +40,10 @@ jobs:
 
   build:
     <<: *job_defaults
+    # We generate a lot of chunks with our build. To improve stability and to reduce
+    # the amount it takes to run, we use a higher resource class with better VM specs.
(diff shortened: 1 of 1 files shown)


devversion:
Sure. I'll add a few numbers here. There aren't a lot since we added the new chunks with the latest commit in master. Builds before this PR (with the new chunks):

* https://circleci.com/gh/angular/material.angular.io/731 ~8min build
* https://circleci.com/gh/angular/material.angular.io/725 ~4min build
* https://circleci.com/gh/angular/material.angular.io/720 Failed. Flaky build.

Builds with this PR:

* https://circleci.com/gh/angular/material.angular.io/735 ~1.5min build
* https://circleci.com/gh/angular/material.angular.io/739 ~1.5min build

So if the average build before this PR would take ~6min, the build would be 75% faster. Also agreed with what Jeremy said. We won't run a lot of builds on CI for this repository.

PR #729, opened 2020-03-05
refactor: account for collision avoidance changes to docs-content

devversion:
With https://github.com/angular/components/commit/ce22f1b3444ad006f00cb95f338e6f2e142d559b, we changed how the docs-content
is composed. Instead of always putting all files at the top-level
of the `docs-content/`, files are now nested based on their structure
in `angular/components` repository. This ensures that there no collisions
for such files, and that multiple files with same file can exist in the docs-content.

The docs app needs to be updated to account for these new paths.

### DIFF
--- a/src/app/pages/component-viewer/component-viewer.ts
+++ b/src/app/pages/component-viewer/component-viewer.ts
@@ -116,6 +116,16 @@ export class ComponentOverview extends ComponentBaseView {
   constructor(componentViewer: ComponentViewer, breakpointObserver: BreakpointObserver) {
     super(componentViewer, breakpointObserver);
   }
+
+  getOverviewDocumentUrl(doc: DocItem) {
+    // Use the explicit overview path if specified. Otherwise, compute an overview path based
+    // on the package name and doc item id. Overviews for components are commonly stored in a
+    // folder named after the component while the overview file is named similarly. e.g.
+    //    `cdk#overlay`     -> `cdk/overlay/overlay.md`
+    //    `material#button` -> `material/button/button.md`
+    const overviewPath = doc.overviewPath || `${doc.packageName}/${doc.id}/${doc.id}.html`;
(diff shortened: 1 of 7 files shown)


src/app/pages/component-viewer/component-viewer.ts
```diff
@@ -116,6 +116,16 @@ export class ComponentOverview extends ComponentBaseView {
...
     super(componentViewer, breakpointObserver);
   }
+
+  getOverviewDocumentUrl(doc: DocItem) {
+    // Use the explicit overview path if specified. Otherwise, compute an overview path based
+    // on the package name and doc item id. Overviews for components are commonly stored in a
+    // folder named after the component while the overview file is named similarly. e.g.
+    //    `cdk#overlay`     -> `cdk/overlay/overlay.md`
+    //    `material#button` -> `material/button/button.md`
+    const overviewPath = doc.overviewPath || `${doc.packageName}/${doc.id}/${doc.id}.html`;
```
Splaktar: This comment assumes a little too much. I.e. that a `material/button/button.md` becomes `material/button/button.html` during the processing/publishing of the docs-content. It would be less confusing (at some distant point in the future or for new contributors) if this was mentioned here.

src/app/shared/documentation-items/documentation-items.ts
```diff
@@ -6,12 +6,21 @@ export interface AdditionalApiDoc {
...
 export interface DocItem {
+  /** Id of the doc item. Used in the URL for linking to the doc. */
   id: string;
+  /** Display name of the doc item. */
   name: string;
+  /** Short summary of the doc item. */
   summary?: string;
+  /** Package which contains the doc item. */
   packageName?: string;
+  /** List of examples. */
```
Splaktar: Example names?

PR #730, opened 2020-03-06
Follow up to example-viewer for collision docs-content change

devversion:
The recent changes to account for the docs-content have been merged. Unfortunately though it seems like parts of the changes have been lost (for the highlighted example files) for whatever reason (maybe rebasing).

This follow-up PR adds the remaining parts needed to account for the docs-content restructure. It ensures that the highlighted example files can be loaded in the docs.


devversion:
@Splaktar Updated. Is it clear now?

Splaktar:
Yeah, much more clear. Thank you.

PR #732, opened 2020-03-06
fix: explicitly set an index for each API docs section

### LINKED ISSUE #731: Table of contents sometimes in the wrong order
The testing section sometimes renders above the main API:

image

I can make this happen by doing a hard reload over and over (it's not consistent). @andrewseguin can you take a look?

andrewseguin:
Link sections were inserted in the order Angular rendered them, which I guess is non-deterministic (now?). 

Fixes #731

### DIFF
--- a/src/app/shared/table-of-contents/table-of-contents.ts
+++ b/src/app/shared/table-of-contents/table-of-contents.ts
@@ -104,7 +104,7 @@ export class TableOfContents implements OnInit, AfterViewInit, OnDestroy {
     this._links = [];
   }
 
-  addHeaders(sectionName: string, docViewerContent: HTMLElement) {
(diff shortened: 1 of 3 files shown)


src/app/shared/table-of-contents/table-of-contents.ts
```diff
@@ -104,7 +104,7 @@ export class TableOfContents implements OnInit, AfterViewInit, OnDestroy {
     this._links = [];
   }
 
-  addHeaders(sectionName: string, docViewerContent: HTMLElement) {
+  addHeaders(sectionName: string, docViewerContent: HTMLElement, sectionIndex = 0) {
```
Splaktar: I have a slight preference for making the `sectionIndex` the 2nd argument rather than the 3rd in both of these functions.

PR #745, opened 2020-04-14
add skip navigation link

annieyw:
A soft nav should auto focus on the content and a hard nav should have the next focus element as skip link.
A hard nav is a full page load (e.g. when coming to the app from another site), a soft nav is a route change within the app. These terms are further explained in https://dzone.com/articles/quick-primer-to-single-page-application-spa-and-tr

### DIFF
--- a/src/app/material-docs-app.html
+++ b/src/app/material-docs-app.html
@@ -1,2 +1,5 @@
+<div id="skip-link-wrapper">
+  <a [href]="skipLinkTarget" color="accent">Skip to main content</a>
+</div>
 <app-navbar class="mat-elevation-z6"></app-navbar>
 <router-outlet></router-outlet>
--- a/src/app/material-docs-app.html
+++ b/src/app/material-docs-app.html
@@ -1,5 +1,2 @@
-<div id="skip-link-wrapper">
-  <a [href]="skipLinkTarget" color="accent">Skip to main content</a>
-</div>
 <app-navbar class="mat-elevation-z6"></app-navbar>
 <router-outlet></router-outlet>
--- a/src/app/material-docs-app.scss
+++ b/src/app/material-docs-app.scss
@@ -26,6 +26,10 @@ material-docs-app > guide-viewer {
   overflow-y: auto;
 }
 
+#main-content {
+  outline: none;
+}
+
 @media (max-width: 720px) {
   material-docs-app {
     top: 92px;
--- a/src/app/material-docs-app.scss
+++ b/src/app/material-docs-app.scss
@@ -26,6 +26,28 @@ material-docs-app > guide-viewer {
   overflow-y: auto;
 }
 
+#skip-link-wrapper {
+  z-index: 10;
+  position: fixed;
+  top: 15px;
+  width: 16%;
+  margin: 0 42% 0 42%;
+  text-align: center;
+  padding: 0.2em 0 0.2em 0;
+  background: white;
+  border-radius: 20px;
+  opacity: 0;
+}
+
+#skip-link-wrapper:focus-within {
+  opacity: 1;
+}
+
+#skip-link-wrapper a {
+  text-decoration: none;
+  color: black;
+}
+
 @media (max-width: 720px) {
   material-docs-app {
     top: 92px;
--- a/src/app/material-docs-app.scss
+++ b/src/app/material-docs-app.scss
@@ -26,28 +26,6 @@ material-docs-app > guide-viewer {
   overflow-y: auto;
 }
 
-#skip-link-wrapper {
-  z-index: 10;
-  position: fixed;
-  top: 15px;
-  width: 16%;
-  margin: 0 42% 0 42%;
-  text-align: center;
-  padding: 0.2em 0 0.2em 0;
-  background: white;
-  border-radius: 20px;
-  opacity: 0;
-}
-
-#skip-link-wrapper:focus-within {
-  opacity: 1;
-}
-
-#skip-link-wrapper a {
-  text-decoration: none;
-  color: black;
-}
-
 @media (max-width: 720px) {
   material-docs-app {
     top: 92px;
--- a/src/app/material-docs-app.ts
+++ b/src/app/material-docs-app.ts
@@ -11,8 +11,6 @@ import {GaService} from './shared/ga/ga';
   encapsulation: ViewEncapsulation.None,
 })
 export class MaterialDocsApp {
-  skipLinkTarget: string;
-
   constructor(router: Router, ga: GaService) {
     let previousRoute = router.routerState.snapshot.url;
     router.events
@@ -26,11 +24,6 @@ export class MaterialDocsApp {
         }
         previousRoute = urlAfterRedirects;
         ga.locationChanged(urlAfterRedirects);
-
-        // set skip link
-        if(!router.url.endsWith('#main-content')) {
-          this.skipLinkTarget = router.url + "#main-content";
-        }
       });
   }
 }
--- a/src/app/material-docs-app.ts
+++ b/src/app/material-docs-app.ts
@@ -11,10 +11,10 @@ import {GaService} from './shared/ga/ga';
   encapsulation: ViewEncapsulation.None,
 })
 export class MaterialDocsApp {
+  skipLinkTarget: string;
 
   constructor(router: Router, ga: GaService) {
     let previousRoute = router.routerState.snapshot.url;
-
     router.events
       .pipe(filter((event: Event) => event instanceof NavigationEnd))
       .subscribe((data: Event) => {
@@ -24,9 +24,13 @@ export class MaterialDocsApp {
         if (!isNavigationWithinComponentView(previousRoute, urlAfterRedirects)) {
           resetScrollPosition();
         }
-
         previousRoute = urlAfterRedirects;
         ga.locationChanged(urlAfterRedirects);
+
+        // set skip link
+        if(!router.url.endsWith('#main-content')) {
+          this.skipLinkTarget = router.url + "#main-content";
+        }
       });
   }
 }
--- a/src/app/pages/component-sidenav/component-sidenav.html
+++ b/src/app/pages/component-sidenav/component-sidenav.html
@@ -11,7 +11,7 @@
   <div class="docs-component-sidenav-content">
     <component-page-header (toggleSidenav)="toggleSidenav(sidenav)"></component-page-header>
     <div class="docs-component-sidenav-inner-content">
-      <main class="docs-component-sidenav-body-content" id="main-content" role="main" aria-label="Side navigation bar">
+      <main class="docs-component-sidenav-body-content" id="main-content" aria-label="Main content">
         <!-- If on large screen, menu resides to left of content -->
         <app-component-nav
           *ngIf="(isScreenSmall | async) === false"
--- a/src/app/pages/component-sidenav/component-sidenav.html
+++ b/src/app/pages/component-sidenav/component-sidenav.html
@@ -11,7 +11,7 @@
   <div class="docs-component-sidenav-content">
     <component-page-header (toggleSidenav)="toggleSidenav(sidenav)"></component-page-header>
     <div class="docs-component-sidenav-inner-content">
-      <main class="docs-component-sidenav-body-content">
+      <main class="docs-component-sidenav-body-content" id="main-content" role="main" aria-label="Side navigation bar">
         <!-- If on large screen, menu resides to left of content -->
         <app-component-nav
           *ngIf="(isScreenSmall | async) === false"
--- a/src/app/pages/component-sidenav/component-sidenav.html
+++ b/src/app/pages/component-sidenav/component-sidenav.html
@@ -11,7 +11,7 @@
   <div class="docs-component-sidenav-content">
     <component-page-header (toggleSidenav)="toggleSidenav(sidenav)"></component-page-header>
     <div class="docs-component-sidenav-inner-content">
-      <main class="docs-component-sidenav-body-content" id="main-content" aria-label="Main content">
+      <main class="docs-component-sidenav-body-content">
         <!-- If on large screen, menu resides to left of content -->
         <app-component-nav
           *ngIf="(isScreenSmall | async) === false"
--- a/src/app/pages/component-sidenav/component-sidenav.ts
+++ b/src/app/pages/component-sidenav/component-sidenav.ts
@@ -39,6 +39,7 @@ import {StackBlitzButtonModule} from '../../shared/stack-blitz';
 import {SvgViewerModule} from '../../shared/svg-viewer/svg-viewer';
 import {MatDrawerToggleResult} from '@angular/material/sidenav/drawer';
 import {MatListModule} from '@angular/material/list';
+import {NavigationFocusModule} from '../../shared/navigation-focus/navigation-focus';
 
 // These constants are used by the ComponentSidenav for orchestrating the MatSidenav in a responsive
 // way. This includes hiding the sidenav, defaulting it to open, changing the mode from over to
@@ -210,7 +211,8 @@ const routes: Routes = [ {
     MatSidenavModule,
     StackBlitzButtonModule,
     SvgViewerModule,
-    RouterModule.forChild(routes)
+    RouterModule.forChild(routes),
+    NavigationFocusModule
   ],
   exports: [ComponentSidenav],
   declarations: [ComponentSidenav, ComponentNav],
--- a/src/app/pages/component-viewer/component-api.html
+++ b/src/app/pages/component-viewer/component-api.html
@@ -1,5 +1,5 @@
 <ng-container *ngIf="componentViewer.componentDocItem | async; let docItem">
-  <span class="cdk-visually-hidden" tabindex="-1" #initialFocusTarget>
+  <span class="cdk-visually-hidden" tabindex="-1">
     API for {{docItem?.id}}
   </span>
 
--- a/src/app/pages/component-viewer/component-viewer.ts
+++ b/src/app/pages/component-viewer/component-viewer.ts
@@ -3,7 +3,6 @@ import {CommonModule} from '@angular/common';
 import {
   Component,
   Directive,
-  ElementRef,
   NgModule,
   OnDestroy,
   OnInit,
@@ -74,7 +73,6 @@ export class ComponentViewer implements OnDestroy {
  */
 @Directive()
 export class ComponentBaseView implements OnInit, OnDestroy {
-  @ViewChild('initialFocusTarget') focusTarget: ElementRef;
   @ViewChild('toc') tableOfContents: TableOfContents;
 
   showToc: Observable<boolean>;
@@ -88,8 +86,6 @@ export class ComponentBaseView implements OnInit, OnDestroy {
 
   ngOnInit() {
     this.componentViewer.componentDocItem.pipe(takeUntil(this.destroyed)).subscribe(() => {
-      // 100ms timeout is used to allow the page to settle before moving focus for screen readers.
-      setTimeout(() => this.focusTarget.nativeElement.id = 'main-content', 100);
       if (this.tableOfContents) {
         this.tableOfContents.resetHeaders();
       }
--- a/src/app/pages/component-viewer/component-viewer.ts
+++ b/src/app/pages/component-viewer/component-viewer.ts
@@ -19,6 +19,7 @@ import {DocItem, DocumentationItems} from '../../shared/documentation-items/docu
 import {TableOfContents} from '../../shared/table-of-contents/table-of-contents';
 import {TableOfContentsModule} from '../../shared/table-of-contents/table-of-contents.module';
 import {ComponentPageTitle} from '../page-title/page-title';
+import {NavigationFocusModule} from "../../shared/navigation-focus/navigation-focus";
 
 @Component({
   selector: 'app-component-viewer',
@@ -88,7 +89,10 @@ export class ComponentBaseView implements OnInit, OnDestroy {
   ngOnInit() {
     this.componentViewer.componentDocItem.pipe(takeUntil(this.destroyed)).subscribe(() => {
       // 100ms timeout is used to allow the page to settle before moving focus for screen readers.
-      setTimeout(() => this.focusTarget.nativeElement.focus({preventScroll: true}), 100);
+      setTimeout(() =>
+      {
+        this.focusTarget.nativeElement.id = 'main-content';//.focus({preventScroll: true})
+      }, 100);
       if (this.tableOfContents) {
         this.tableOfContents.resetHeaders();
       }
@@ -163,6 +167,7 @@ export class ComponentExamples extends ComponentBaseView {
     DocViewerModule,
     CommonModule,
     TableOfContentsModule,
+    NavigationFocusModule,
   ],
   exports: [ComponentViewer],
   declarations: [ComponentViewer, ComponentOverview, ComponentApi, ComponentExamples],
--- a/src/app/pages/component-viewer/component-viewer.ts
+++ b/src/app/pages/component-viewer/component-viewer.ts
@@ -19,7 +19,7 @@ import {DocItem, DocumentationItems} from '../../shared/documentation-items/docu
 import {TableOfContents} from '../../shared/table-of-contents/table-of-contents';
 import {TableOfContentsModule} from '../../shared/table-of-contents/table-of-contents.module';
 import {ComponentPageTitle} from '../page-title/page-title';
-import {NavigationFocusModule} from "../../shared/navigation-focus/navigation-focus";
+import {NavigationFocusModule} from '../../shared/navigation-focus/navigation-focus';
 
 @Component({
   selector: 'app-component-viewer',
@@ -89,10 +89,7 @@ export class ComponentBaseView implements OnInit, OnDestroy {
   ngOnInit() {
     this.componentViewer.componentDocItem.pipe(takeUntil(this.destroyed)).subscribe(() => {
       // 100ms timeout is used to allow the page to settle before moving focus for screen readers.
-      setTimeout(() =>
-      {
-        this.focusTarget.nativeElement.id = 'main-content';
-      }, 100);
+      setTimeout(() => this.focusTarget.nativeElement.id = 'main-content', 100);
       if (this.tableOfContents) {
         this.tableOfContents.resetHeaders();
       }
--- a/src/app/pages/component-viewer/component-viewer.ts
+++ b/src/app/pages/component-viewer/component-viewer.ts
@@ -91,7 +91,7 @@ export class ComponentBaseView implements OnInit, OnDestroy {
       // 100ms timeout is used to allow the page to settle before moving focus for screen readers.
       setTimeout(() =>
       {
-        this.focusTarget.nativeElement.id = 'main-content';//.focus({preventScroll: true})
+        this.focusTarget.nativeElement.id = 'main-content';
       }, 100);
       if (this.tableOfContents) {
         this.tableOfContents.resetHeaders();
--- a/src/app/pages/guide-list/guide-list.html
+++ b/src/app/pages/guide-list/guide-list.html
@@ -1,8 +1,8 @@
 <header class="docs-primary-header">
-  <h1>Guides</h1>
+  <h1 focusOnNavigation>Guides</h1>
 </header>
 
-<main>
+<main id="main-content" role="main" aria-label="Guide list">
   <mat-nav-list class="docs-guide-list">
     <a mat-list-item *ngFor="let guide of guideItems.getAllItems()"
         class="docs-guide-item"
--- a/src/app/pages/guide-list/guide-list.html
+++ b/src/app/pages/guide-list/guide-list.html
@@ -1,8 +1,8 @@
 <header class="docs-primary-header">
-  <h1 focusOnNavigation>Guides</h1>
+  <h1>Guides</h1>
 </header>
 
-<main id="main-content" role="main" aria-label="Guide list">
+<main focusOnNavigation id="main-content" aria-label="Guide list">
   <mat-nav-list class="docs-guide-list">
     <a mat-list-item *ngFor="let guide of guideItems.getAllItems()"
         class="docs-guide-item"
--- a/src/app/pages/homepage/homepage.html
+++ b/src/app/pages/homepage/homepage.html
@@ -1,5 +1,5 @@
-<header class="docs-header-background"
-        [class.is-next-version]="isNextVersion" id="main-content" role="main" aria-label="Get started">
+<header focusOnNavigation class="docs-header-background"
+        [class.is-next-version]="isNextVersion" id="main-content" aria-label="Get started">
   <div class="docs-header-section">
     <div class="docs-header-headline">
       <h1 class="mat-h1">Angular Material</h1>
--- a/src/app/pages/homepage/homepage.html
+++ b/src/app/pages/homepage/homepage.html
@@ -1,5 +1,5 @@
 <header class="docs-header-background"
-        [class.is-next-version]="isNextVersion">
+        [class.is-next-version]="isNextVersion" id="main-content" role="main" aria-label="Get started">
   <div class="docs-header-section">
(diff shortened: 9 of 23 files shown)


atscott:
I think the headbutting headers was merged yesterday, right? Can you drop that commit from this PR and update the commit message for the second one to be more descriptive?

src/app/pages/component-sidenav/component-sidenav.html
```diff
@@ -11,7 +11,7 @@
   <div class="docs-component-sidenav-content">
     <component-page-header (toggleSidenav)="toggleSidenav(sidenav)"></component-page-header>
     <div class="docs-component-sidenav-inner-content">
-      <main class="docs-component-sidenav-body-content">
+      <main class="docs-component-sidenav-body-content" id="main-content" role="main" aria-label="Side navigation bar">
```
atscott, line 14: Is this the correct label? It looks like this is the main body contents, not the navigation bar.

src/app/material-docs-app.html
```diff
@@ -1,2 +1,5 @@
+<div id="skip-link-wrapper">
```
mmalerba, line 1: We like to uses classes over ids when possible because the same class can appear on more than 1 element and CSS rules targeted at classes are less specific than rules targeted at ids.

src/app/shared/navigation-focus/navigation-focus.ts
```diff
@@ -9,16 +11,32 @@ let lastTimeoutId = -1;
...
-    lastTimeoutId = window.setTimeout(() => this.el.nativeElement.focus({preventScroll: true}),
-      100);
+    let previousUrl = window.location.hostname;
+    this.router.events
+      .pipe(filter((event: Event) => event instanceof NavigationEnd))
+      .subscribe(() => {
+        const afterUrl = window.location.hostname;
+        // only want to focus on the element if it the navigation was a soft nav,
+        // otherwise first focusable element should be the skip link
+        if (isSoftNav(previousUrl)) {
```
atscott, line 24: Can you determine this from the `id` on the `RouterEvent` https://angular.io/api/router/RouterEvent? A hard nav would have an id of 0, right?

src/app/material-docs-app.scss
```diff
@@ -26,6 +26,28 @@ material-docs-app > guide-viewer {
   overflow-y: auto;
 }
 
+#skip-link-wrapper {
+  z-index: 10;
+  position: fixed;
```
mmalerba, line 31: Should this be moved inside the `app-navbar`? That would probably make it so we don't need the `position: fixed` and `z-index`, since the app bar already has fixed positioning

src/app/material-docs-app.scss
```diff
@@ -26,6 +26,28 @@ material-docs-app > guide-viewer {
...
+  top: 15px;
+  width: 16%;
+  margin: 0 42% 0 42%;
+  text-align: center;
+  padding: 0.2em 0 0.2em 0;
+  background: white;
+  border-radius: 20px;
+  opacity: 0;
```
mmalerba, line 39: Is `opacity: 0` sufficient to prevent the element from receiving click events?

src/app/material-docs-app.ts
```diff
@@ -24,9 +24,13 @@ export class MaterialDocsApp {
         if (!isNavigationWithinComponentView(previousRoute, urlAfterRedirects)) {
           resetScrollPosition();
         }
-
         previousRoute = urlAfterRedirects;
         ga.locationChanged(urlAfterRedirects);
+
+        // set skip link
+        if(!router.url.endsWith('#main-content')) {
+          this.skipLinkTarget = router.url + "#main-content";
```
mmalerba, line 32: I could see this breaking for a URL like `.../stuff#not-the-main-content`. I think you should just be able to set the `href` to `#main-content` and the rest of the URL is assumed

src/app/pages/guide-list/guide-list.html
```diff
@@ -1,8 +1,8 @@
 <header class="docs-primary-header">
-  <h1>Guides</h1>
+  <h1 focusOnNavigation>Guides</h1>
```
mmalerba, line 2: It seems weird to me that the auto-focused element and the element the skip link takes you to are not the same

src/app/pages/homepage/homepage.html
```diff
@@ -1,5 +1,5 @@
 <header class="docs-header-background"
-        [class.is-next-version]="isNextVersion">
+        [class.is-next-version]="isNextVersion" id="main-content" role="main" aria-label="Get started">
```
mmalerba, line 2: On other pages the `<main>` element is the `#main-content`, is this a mistake?

src/app/shared/navigation-focus/navigation-focus.ts
```diff
@@ -9,16 +11,32 @@ let lastTimeoutId = -1;
...
+          lastTimeoutId = window.setTimeout(() => this.el.nativeElement.focus({preventScroll: true}),
+            100);
+        }
+        previousUrl = afterUrl;
+      });
   }
 }
 
+function isSoftNav(url: string) {
+  const withinApp = /(localhost|material.angular.io)/;
```
mmalerba, line 36: This is brittle, I would just compare the old and new hostname with `===`.

@atscott do you know if there's a more "Angular" way of distinguishing soft nav from hard nav?

src/app/shared/navigation-focus/navigation-focus.ts
```diff
@@ -9,16 +11,32 @@ let lastTimeoutId = -1;
...
 
-  constructor(private el: ElementRef) {}
+  constructor(private el: ElementRef, private router: Router) {}
 
   ngOnInit() {
-    clearTimeout(lastTimeoutId);
-    // 100ms timeout is used to allow the page to settle before moving focus for screen readers.
-    lastTimeoutId = window.setTimeout(() => this.el.nativeElement.focus({preventScroll: true}),
-      100);
+    let previousUrl = window.location.hostname;
```
mmalerba, line 17: Technically `previousHostName` which is not the same as a URL

src/app/shared/navigation-focus/navigation-focus.ts
```diff
@@ -9,16 +11,32 @@ let lastTimeoutId = -1;
...
+    this.router.events
+      .pipe(filter((event: Event) => event instanceof NavigationEnd))
+      .subscribe(() => {
+        const afterUrl = window.location.hostname;
```
mmalerba, line 21: It looks like this variable is only used once, might as well just inline it

src/app/shared/navigation-focus/navigation-focus.ts
mmalerba: I figured there had to be some way, but I had no idea what it was. Also, I would never guess that that's what 0 means from the docs

atscott: Yea, it's not immediately apparent. Using the `id` won't work if the initial navigation results in a redirect from a guard calling `router.navigate` or returning `UrlTree`. The initial navigation would then be cancelled and the following would have `id` 1... Still trying to think of a more "Angular" way to do this that works in all cases. 
I'm not 100% confident that the current way would work in the redirect case either. @vxlydia - do you think you can you test this out?

src/app/pages/homepage/homepage.html
vxlydia: I wasn't sure whether to put it here or in main because there are 2 "Get Started" buttons and this one was the first one

src/app/shared/navigation-focus/navigation-focus.ts
vxlydia: Changed to using `id` from `RouterEvent`

src/app/shared/navigation-focus/navigation-focus.ts
vxlydia: I tried checking if `id` is 0 and it seems so work. I'm not sure how to test for the redirect case, could you point me to an example?

atscott: The only guard I could find was here: https://github.com/angular/material.angular.io/blob/master/src/app/pages/component-sidenav/component-sidenav-can-load-guard.ts and it does call `router.navigate`. So to test out this cancellation/redirect behavior, navigate to `localhost:4200/nonsenseroute`.

src/app/pages/component-sidenav/component-sidenav.html
Splaktar: A `main` element also doesn't need to have an additional `role="main"` defined.

Splaktar:
Can you please add a link to the PR description to a site/blog/documentation that defines and describes this soft-nav and hard-nav pattern and the vocabulary involved? Maybe there's a WAI-ARIA practice that covers this?

src/app/pages/homepage/homepage.html
Splaktar: It doesn't seem expected or ideal for a `header` element to have `role="main"`.

src/app/shared/navigation-focus/navigation-focus.ts
Splaktar: We'll want this to work with the following as well
- https://next.material.angular.io/
- https://v6.material.angular.io/ (also v5, v7, v8, etc)

src/app/shared/navigation-focus/navigation-focus.ts
```diff
@@ -9,16 +11,32 @@ let lastTimeoutId = -1;
...
+        // only want to focus on the element if it the navigation was a soft nav,
```
Splaktar, line 22:
```suggestion
        // only want to focus on the element if the navigation was a soft nav,
```

mmalerba:
@Splaktar This post talks about it a bit: https://dzone.com/articles/quick-primer-to-single-page-application-spa-and-tr

TL;DR a hard nav is a full page load (e.g. when coming to the app from another site), a soft nav is a route change within the app. 

I couldn't really find anything talking about how a skip link should work with soft nav, but we thought it makes sense to put the focus at the top on the first page load so people can explore the nav links, and then put the focus at the content the user is trying to navigate to on subsequent route changes

src/app/material-docs-app.scss
mmalerba: Yeah just checked this, `opacity:0` is not sufficient. We have a class in the CDK called `cdk-visually-hidden` that should handle this, I would just apply that class to the element

src/app/pages/homepage/homepage.html
mmalerba: Ok this makes sense after I looked at the page. Like Michael said, I would just remove the `role="main"`

Splaktar:
OK, that makes sense. 

On https://material.angularjs.org (AngularJS Material), on a hard nav we don't have any focus, in this case the 2nd tabbable element is the skip link and the third is the nav.

For soft nav, we focus the header text in the toolbar (i.e. "Demos > Checkbox", [Checkbox Demo link](https://material.angularjs.org/latest/demo/checkbox)) to allow the user to quickly get into the content.

src/app/shared/navbar/navbar.html
```diff
@@ -1,6 +1,11 @@
 <!-- TODO: figure out if the <nav> should go inside of a <header> element. -->
 <nav class="docs-navbar-header" aria-label="Top Toolbar"
      [class.is-next-version]="isNextVersion">
+  <div class="skip-link-wrapper" [class.cdk-visually-hidden]=skipLinkHidden>
+    <a [href]=skipLinkTarget id="skip-link" (focus)="setSkipLinkVisible()" (blur)="setSkipLinkInvisible()">
```
mmalerba, line 5: is the ID needed for some reason? I don't see any code that references it

mmalerba, line 5: nit: rather than having 2 method you could just have one that takes an argument, or 0 methods: `(focus)="skipLinkHidden = false"`

src/app/shared/navbar/navbar.ts
```diff
@@ -19,6 +20,17 @@ const SECTIONS_KEYS = Object.keys(SECTIONS);
...
   isNextVersion = location.hostname.startsWith('next.material.angular.io');
+  skipLinkTarget: string;
+  skipLinkHidden = true;
+  constructor(router: Router) {
+    router.events
+      .pipe(filter((event: Event) => event instanceof NavigationEnd))
+      .subscribe(() => {
+        // set skip link
+        const baseUrl = router.url.split('#')[0];
+        this.skipLinkTarget = `${baseUrl}#main-content`;
```
mmalerba, line 31: can we call this `skipLinkHref`, `target` is actually a different attribute on `<a>` tags that tells it whether to open in the same window or a new one

src/app/material-docs-app.ts
vxlydia: setting `href` directly to `#main-content` will cause the page to redirect to the homepage. 
a similar issue is addressed in [doc-viewer.ts](https://github.com/angular/material.angular.io/blob/master/src/app/shared/doc-viewer/header-link.ts)

src/app/shared/navigation-focus/navigation-focus.ts
```diff
@@ -9,16 +11,29 @@ let lastTimeoutId = -1;
...
-    clearTimeout(lastTimeoutId);
-    // 100ms timeout is used to allow the page to settle before moving focus for screen readers.
-    lastTimeoutId = window.setTimeout(() => this.el.nativeElement.focus({preventScroll: true}),
-      100);
+    this.router.events
+      .pipe(filter((event: Event): event is RouterEvent => event instanceof RouterEvent))
+      .subscribe((routerEvent: RouterEvent) => {
+        // only want to focus on the element if the navigation was a soft nav,
+        // otherwise first focusable element should be the skip link
+        if (!(routerEvent.id === 0)) {
```
mmalerba, line 22: can we make this `if(isSoftNav(routerEvent))`, easier to follow the logic that way

src/app/shared/navbar/navbar.html
```diff
@@ -1,6 +1,11 @@
 <!-- TODO: figure out if the <nav> should go inside of a <header> element. -->
 <nav class="docs-navbar-header" aria-label="Top Toolbar"
      [class.is-next-version]="isNextVersion">
+  <div class="skip-link-wrapper" [class.cdk-visually-hidden]=skipLinkHidden>
```
atscott, line 4: Style nit: always use quotes for the attribute values, i.e. `[class.cdk-visually-hidden]="skipLinkHidden"` and `[href]="skipLinkTarget"`

src/app/shared/navbar/navbar.html
vxlydia: added that plus `(blur)="skipLinkHidden = true"` otherwise the skip link would stay there

src/app/shared/navigation-focus/navigation-focus.ts
```diff
@@ -9,16 +11,34 @@ let lastTimeoutId = -1;
...
+  private currentLocation: URL;
+
+  constructor(private el: ElementRef, private router: Router) {}
 
   ngOnInit() {
-    clearTimeout(lastTimeoutId);
-    // 100ms timeout is used to allow the page to settle before moving focus for screen readers.
-    lastTimeoutId = window.setTimeout(() => this.el.nativeElement.focus({preventScroll: true}),
-      100);
+    this.currentLocation = new URL(window.location.href);
```
atscott, line 20: I'm confused about how this works. Wouldn't the location *have* to be `localhost`/`material.io` at this point? I don't know if you're able to detect "hard" navs with this method. Maybe I'm missing something though.

atscott: There's already code in the `MaterialDocsApp` that tracks `NavigationEnd` events. Can you maybe add another that's something like:
```
router.events
      .pipe(
                filter((event: Event) => event instanceof NavigationEnd)),
                skip(1),
                take(1),
      ).subscribe(() => globalState.internallyNavigated = true);
```

src/app/shared/navigation-focus/navigation-focus.ts
```diff
@@ -9,16 +11,34 @@ let lastTimeoutId = -1;
...
+    this.router.events.pipe(filter((event: Event) => event instanceof NavigationEnd))
+      .subscribe(() => {
+        this.previousLocation = this.currentLocation;
+        this.currentLocation = new URL(window.location.href);
+        if (this.currentLocation.hash) {
+          clearTimeout(lastTimeoutId);
+          // 100ms timeout is used to allow the page to settle before moving focus for screen readers.
+          lastTimeoutId = window.setTimeout(() => document.getElementById(this.currentLocation.hash)!.focus(), 100);
```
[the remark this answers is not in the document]
mmalerba: I think we had some problems at some point in the past where the element with the ID in question was late loaded and so it just linked to the top. We can test commenting this out and see if links within the component overview still work

src/app/shared/navigation-focus/navigation-focus.ts
mmalerba: Does the router not fire events when I navigate from e.g. about:blank --> m.a.io/components/autocomplete/overview? I was under the impression that it did, but I admittedly don't know much about the router

src/app/shared/navigation-focus/navigation-focus.ts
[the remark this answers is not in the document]
atscott: Ahhh, I see. That could be the case.

src/app/shared/navigation-focus/navigation-focus.ts
atscott: It does fire events for the initial navigation. about:blank --> m.a.io/components/autocomplete/overview would fire all the navigation events. I don't think I'm answering the question you're asking though... By the time any JS runs on the page, the URL is already m.a.io/components/autocomplete/overview

mmalerba: Oh, yeah I see what you're saying. I think for this method to work properly, the previousUrl would have to be globally saved and initially null, then we update it after each navigation. I think Annie is planning to switch this back to using the id property to detect soft navs like you originally suggested anyways

src/app/shared/navigation-focus/navigation-focus.ts
```diff
@@ -1,24 +1,57 @@
...
+  constructor(private el: ElementRef, private router: Router) {
+    // subscribing to events in constructor to catch the correct NavigationEnd/ResolveEnd from previous page
+    // and determine whether the element in question needs to be focused
+    // switching between different components/cdk produces a ResolveEnd instead of NavigationEnd event
+    // the focus is set in ngOnInit to account for late loaded content
+    this.subscriptions.add(
+      this.router.events
+        .pipe(
+          filter((event: Event) => event instanceof NavigationEnd || event instanceof ResolveEnd),
+          map(navEnd => navEnd as NavigationEnd))
```
mmalerba, line 27: This map is unnecessary, it's just serving as an expensive type cast. Also it seems like maybe a lie since it could be a `ResolveEnd` - why do we want `ResolveEnd` in there? That seems not right

You could maybe try using this alternate signature of filter (notice 2 type parameters):
```ts
function filter<T, S extends T>(predicate: (value: T, index: number) => value is S,
                                       thisArg?: any): OperatorFunction<T, S>;
```

In the context of your code it would look something like:
```ts
filter<RouterEvent, NavigationEnd>(...)
```

src/app/shared/navigation-focus/navigation-focus.ts
```diff
@@ -1,24 +1,57 @@
...
-export class NavigationFocus implements OnInit {
+export class NavigationFocus implements OnInit, OnDestroy {
   @HostBinding('tabindex') role = '-1';
 
-  constructor(private el: ElementRef) {}
+  private softNavFocus = false;
+  private subscriptions = new Subscription();
+
+  constructor(private el: ElementRef, private router: Router) {
+    // subscribing to events in constructor to catch the correct NavigationEnd/ResolveEnd from previous page
```
mmalerba, line 19: I would change this comment to:
```
// We need to subscribe in the constructor in order to catch the `NavigationEnd` event 
// from navigating from the previous page to this page. This event fires before 
// `ngOnInit` is called. However we must wait until `ngOnInit` for the element to be 
// rendered before we can focus it
```

src/app/shared/navigation-focus/navigation-focus.ts
```diff
@@ -1,24 +1,57 @@
...
+      this.router.events
+        .pipe(
+          filter((event: Event) => event instanceof NavigationEnd || event instanceof ResolveEnd),
+          map(navEnd => navEnd as NavigationEnd))
+      .subscribe((navEnd: NavigationEnd) => {
+        const currentLocation = new URL(window.location.href);
+        if (!currentLocation.hash && isSoftNav(navEnd)) {
+          this.softNavFocus = true;
+          // navigating between CDK/Components pages only runs ngOnInit once so the focus must also be set here
+          this.el.nativeElement.focus({preventScroll: true});
```
mmalerba, line 33: I thought we were waiting until `ngOnInit` to focus?

src/app/shared/navigation-focus/navigation-focus.ts
```diff
@@ -1,24 +1,57 @@
...
 
   ngOnInit() {
-    clearTimeout(lastTimeoutId);
-    // 100ms timeout is used to allow the page to settle before moving focus for screen readers.
-    lastTimeoutId = window.setTimeout(() => this.el.nativeElement.focus({preventScroll: true}),
-      100);
+    if (this.softNavFocus) {
+      // 100ms timeout is used to allow the page to settle before moving focus for screen readers.
+      clearTimeout(lastTimeoutId);
+      lastTimeoutId = window.setTimeout(()=>this.el.nativeElement.focus({preventScroll: true}), 100);
```
mmalerba, line 42: I wonder if we could eliminate the `setTimeout` by doing the focus in `ngAfterViewInit` instead. Might be worth testing

src/app/shared/navigation-focus/navigation-focus.ts
```diff
@@ -1,24 +1,57 @@
...
+    }
+  }
+
+  ngOnDestroy(): void {
```
mmalerba, line 46: nit: don't need to explicitly put `: void`

src/app/shared/navigation-focus/navigation-focus.ts
```diff
@@ -1,24 +1,57 @@
...
+    this.subscriptions.unsubscribe();
   }
 }
 
+function isSoftNav(navigationEnd: NavigationEnd) {
+  return navigationEnd.id !== 1;
```
mmalerba, line 52: I thought @atscott said `0` was a hardnav? How did you get `1`, is that just from experimenting?

src/app/shared/navbar/navbar.html
```diff
@@ -1,6 +1,11 @@
 <!-- TODO: figure out if the <nav> should go inside of a <header> element. -->
 <nav class="docs-navbar-header" aria-label="Top Toolbar"
      [class.is-next-version]="isNextVersion">
+  <div class="skip-link-wrapper" [class.cdk-visually-hidden]="skipLinkHidden" >
```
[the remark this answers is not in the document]
atscott: I tried investigating this as well and couldn't find a way to make `routerLink` work. In general, that would be more desirable, but because the `NavBar` exists at the root, I couldn't figure out how to get the navigation to be relative to the final activated component rather than the root. Otherwise, you could use `<a [routerLink]="[]" fragment="main-content"></a>` or something similar to update the fragment.

src/app/shared/navigation-focus/navigation-focus.ts
vxlydia: I confirmed with him and he said 0 is actually filtered out so it should be 1
https://github.com/angular/angular/blob/master/packages/router/src/router.ts#L448

atscott: Yea, sorry about that. The `initialNavigation` would be `1`, as the `id` is incremented from `0` to `1` before calling `transitions.next`: https://github.com/angular/angular/blob/master/packages/router/src/router.ts#L1151. To be honest, I don't even know if that filter is necessary in the router pipe (it may have been at some point in the past).

mmalerba: Ah ok, do you think this is something worth having a more official API for? If so maybe @vxlydia can send a PR for it

src/app/pages/component-viewer/component-api.html
```diff
@@ -1,5 +1,5 @@
 <ng-container *ngIf="componentViewer.componentDocItem | async; let docItem">
-  <span class="cdk-visually-hidden" tabindex="-1" #initialFocusTarget>
+  <span class="cdk-visually-hidden" tabindex="-1" #initialFocusTarget focusOnNavigation>
```
atscott, line 2: Can we remove this and use `@ViewChild(FocusOnNavigation)` instead in `ComponentBaseView`?

src/app/shared/navigation-focus/navigation-focus.ts
atscott: That could be useful. Maybe something simple like: "each navigation has a unique id and is available on the RouterEvents" and potentially something about how the `id` is a number that's incremented. I'd be wary of making any promises about the meaning of the id beyond that.

src/app/shared/navigation-focus/navigation-focus.ts
vxlydia: navigating between the CDK/Components page only runs ngAfterViewInit once because it's the same component (I believe) so we need to track navigating between them through the router events

src/app/pages/component-sidenav/component-sidenav.ts
```diff
@@ -39,6 +39,7 @@ import {StackBlitzButtonModule} from '../../shared/stack-blitz';
 import {SvgViewerModule} from '../../shared/svg-viewer/svg-viewer';
 import {MatDrawerToggleResult} from '@angular/material/sidenav/drawer';
 import {MatListModule} from '@angular/material/list';
+import {NavigationFocusModule} from "../../shared/navigation-focus/navigation-focus";
```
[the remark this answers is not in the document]
vxlydia: I figured it out :)

src/app/shared/navigation-focus/navigation-focus.ts
vxlydia: Seemed to work. Removed the relevant vars for timeout

src/app/pages/component-viewer/component-api.html
vxlydia: I'm not sure if this is what you meant but  I removed the focusOnNavigation directives on the api/over/example pages and now just have 1 on the base component viewer

src/app/shared/navigation-focus/navigation-focus.ts
vxlydia: Removed map and changed to just using `ResolveEnd`

src/app/pages/component-viewer/component-api.html
atscott: Oh, that's a good change as well! What I meant was to remove the `#initialFocusTarget` and change `@ViewChild('initialFocusTarget') focusTarget: ElementRef;` in `ComponentBaseView` to `@ViewChild(FocusOnNavigation, {read: ElementRef}) focusTarget: ElementRef;` since we have a directive now that we can use as the selector and don't really need the 'initialFocusTarget' ref. It's not really necessary though, just a nit.

src/app/shared/navigation-focus/navigation-focus.ts
mmalerba: Yeah that was kind of my point, making an assumption about the id _is_ hacky, you already pointed out that the filter might be unnecessary and if we remove it we'd need to change this check to compare to `0` instead of `1`. I was thinking more like adding a property to the event that can be more reliably checked against in this type of situation (I'm sure there's many ways we could accomplish this, but for example we could add `{from: Route|null, to: Router|null}` - that way a simple `if(event.from)` would tell us if its a softnav)

atscott: Oh, sorry - I read that as "official docs" rather than "official API". Yes, there are a lot of holes in the router APIs but I can't necessarily recommend spending too much time on new features since the plan for the future of the router is unclear. Adding more API surface that we have to maintain through a potential refactor may be undesirable.

That said, I would absolutely file an issue for this and I can triage it as a feature request. Then we can also see if it gains community support.

src/app/shared/navigation-focus/navigation-focus.ts
vxlydia: Actually it didn't work for this specific case (nav from non cdk/component page to cdk/component page). Kept setting the focus in ngOnInit with timeout

vxlydia:
navigation between pages should be functional now; PTAL

mmalerba:
Looking pretty good, couple things I noticed on the demo:
- There's an ugly outline on the component that's focused after following the skip-link. I think you can remove it with this CSS:
`#main-content { outline: none; }`
- When you navigate to any of the guides (e.g. http://localhost:4200/guide/getting-started) the skiplink doesn't work because there is nothing with the id `main-content`. This is the fragility I mentioned in one of the comments above. We can come up with a more robust solution in a future PR, for now lets just make sure these pages have an element with the right id.
- When I follow the skiplink on http://localhost:4200/components/autocomplete/overview it takes me past the tab headers, it should probalby land just before the tab headers so I can go to the overview, api, or examples easily

src/app/pages/component-viewer/component-viewer.ts
```diff
@@ -88,7 +89,7 @@ export class ComponentBaseView implements OnInit, OnDestroy {
   ngOnInit() {
     this.componentViewer.componentDocItem.pipe(takeUntil(this.destroyed)).subscribe(() => {
       // 100ms timeout is used to allow the page to settle before moving focus for screen readers.
-      setTimeout(() => this.focusTarget.nativeElement.focus({preventScroll: true}), 100);
+      setTimeout(() => this.focusTarget.nativeElement.id = 'main-content', 100);
```
mmalerba, line 92: We probably don't need the setTimeout since we're not triggering a focus anymore, should be fine to set the id immediately

src/app/shared/navbar/navbar.ts
```diff
@@ -17,8 +19,21 @@ const SECTIONS_KEYS = Object.keys(SECTIONS);
...
+  private subscriptions = new Subscription();
   isNextVersion = location.hostname.startsWith('next.material.angular.io');
+  skipLinkHref: string;
+  skipLinkHidden = true;
+
+  constructor(router: Router) {
+    this.subscriptions.add(router.events
+      .pipe(filter((event: Event) => event instanceof NavigationEnd))
+      .subscribe(() => {
+        // set skip link
```
mmalerba, line 32: comment not useful, improve or remove

src/app/shared/navbar/navbar.ts
```diff
@@ -17,8 +19,21 @@ const SECTIONS_KEYS = Object.keys(SECTIONS);
   templateUrl: './navbar.html',
   styleUrls: ['./navbar.scss']
 })
-export class NavBar {
+export class NavBar implements OnDestroy {
+  private subscriptions = new Subscription();
```
atscott, line 23: A common approach for unsubscribing is to have an onDestroy observable, i.e.
```
private readonly onDestroy = new Subject<void>();

...
router.events.pipe(takeUntil(this.onDestroy), ...).subscribe(...);
...

ngOnDestroy() {
  this.onDestroy.next();
  this.onDestroy.complete();
}
```

Up to you - this is more or less an FYI. It's more useful when you have multiple subscriptions that you need to unsubscribe from or as a project-wide convention.

atscott: I misread how this was being used. I like this `subscriptions.add` too :)

src/app/pages/component-viewer/component-viewer.ts
annieyw: I was able to just remove this entirely because now `main-content` is set to the the tab headers.

annieyw:
- remove focus outline 
- add `main-content` to individual guides page
- adjust skip link to land on tabs for individual component/cdk pages to allow easy nav to overview/api/examples

PTAL

Issue #744, opened 2020-04-14
Migrate from node-sass to sass


Splaktar:
We currently use `node-sass` in `build-themes.sh`. It doesn't support the new module system and we are generally moving towards using the `sass` package in other repositories. We need to make that move here as well.

Issue #743, opened 2020-04-13
Suggestion for pre-built themes


vazmin:
Building themes through the shell is not friendly enough for window users.

I try to make a small modification to the angular.json, like this:

```json
"options": {
    "extractCss": true,
    "styles": [{
        "lazy": true,
        "inject": false,
        "input": "src/assets/custom-themes/pink-bluegrey.scss",
        "bundleName": "assets/pink-bluegrey"
    }]
}
```
This is more versatile and plays the same role. One theme, one file, lazy loading.

And you can import the theming this way 
`@import '~@angular/material/theming';`, 
not this way 
`@import "../../../../node_modules/@angular/material/theming";`

Splaktar:
@vazmin you might want to take a look at [this blog post](https://dev.to/angular/angular-apps-with-a-theme-switch-like-the-ngular-material-site-darkmode-3mno) which is based on our implementation, but makes a few changes.

Splaktar:
As for building and developing this project, the Bash shell is required. Bash is well supported on Windows in a variety of flavors (Cygwin, Git Bash, WSL, and more).

As for using `node-sass`, we do have plans to migrate this to use the `sass` package in the future.

vazmin:
> @vazmin you might want to take a look at [this blog post](https://dev.to/angular/angular-apps-with-a-theme-switch-like-the-ngular-material-site-darkmode-3mno) which is based on our implementation, but makes a few changes.

@Splaktar Thanks, that's a great blog post.


> As for building and developing this project, the Bash shell is required. Bash is well supported on Windows in a variety of flavors (Cygwin, Git Bash, WSL, and more).

I am using wsl, he is very good, but the implementation of `bash npm --version` does not work, and export path is required. So I use `angular-cli` instead of `build-themes.sh`

PR #746, opened 2020-04-15
Pre-build themes using Angular CLI (sass + sass-loader) instead of node-sass Bash script

vazmin:
1. Use `angular-cli` instead of directly using `node-sass`
2. Use `@import '~'`

#743 #744


vazmin:
@googlebot I signed it!

yarn.lock
```diff
@@ -211,10 +211,6 @@
   version "9.2.1-sha-b0424a351"
   resolved "https://codeload.github.com/angular/material2-docs-content/tar.gz/66f276a9526a373b4c252370514f10de130561f0"
 
-"@angular/components-examples@angular/material2-docs-content#master":
```
Splaktar, line 214: This shouldn't be removed. It looks like you rebuilt the entire lock file? If so, then you'll need to run `npm run build:content`.

Splaktar:
Thanks for proposing these changes. They look interesting, but I need to take a look at them locally to verify how they work.

yarn.lock
vazmin: I try run 'yarn run build:content' and 'yarn upgrade @angular/components-examples', but `yarn.lock` has not changed

Splaktar: AH OK, that makes sense. The diff didn't show that the one above it was for the same package.

vazmin:
> The only difference I see is that the lazily loaded pre-built theme `.css` files are about `0.2 kB` larger. 

This is true in the case of `yarn run build`:

using node-sass
```css
.mat-ripple-element {
  background-color: rgba(0, 0, 0, 0.1); }
```

using angular cli
```css
.mat-ripple-element {
  background-color: rgba(0, 0, 0, 0.1);
}
```
An extra line break.

`yarn run prod-build` is the same. 

but now the site is using [unoptimized css](https://material.angular.io/assets/deeppurple-amber.css) instead of [optimized css](https://material.angular.io/deeppurple-amber.css).

Of course I also think that is negligible.

jelbourn:
@vazmin to confirm, the prod mode CSS output it identical, and the dev mode CSS preserves whitespace?

vazmin:
@jelbourn 

> the dev mode CSS preserves whitespace?

yes， the default options of `output-style  (nested | expanded | compact | compressed)`  are different

angular cli: 
[ outputStyle: 'expanded',](https://github.com/angular/angular-cli/blob/254994db85398d847d4ca6608c4019fc1ea7a583/packages/angular_devkit/build_angular/src/angular-cli-files/models/webpack-configs/styles.ts#L148)

node-sass:
[ 'output-style': 'nested',](https://github.com/angular/angular-cli/blob/254994db85398d847d4ca6608c4019fc1ea7a583/packages/angular_devkit/build_angular/src/angular-cli-files/models/webpack-configs/styles.ts#L148)

> to confirm, the prod mode CSS output it identical

build.configurations.production.optimization is true

- - -

code output before modification：
```
"assets": [
   {
      "glob": "**/*",
      "input": "src/assets",
      "output": "/assets"
   }
]
```
will output
`dist/assets/indigo-pink.css`  (unoptimized css, the site requested this)

```
{
  "inject": false,
  "input": "src/assets/indigo-pink.css",
  "bundleName": "indigo-pink"
}
```
will output
`dist/indigo-pink.css` (optimized css)

- - -

code output before modification：
```
{
  "lazy": true,
  "inject": false,
  "input": "src/styles/custom-themes/indigo-pink.scss",
  "bundleName": "assets/indigo-pink"
}
```
will output
`dist/assets/indigo-pink.css`  (optimized css)

The `intellij idea` I use has the function` Compare with` which provides a comparison file, the content of the optimized css is the same,

angular.json
```diff
@@ -21,6 +21,7 @@
             "index": "src/index.html",
             "main": "src/main.ts",
             "tsConfig": "tsconfig.app.json",
+            "extractCss": true,
```
devversion, line 24: We could remove `extractCss` from the production configuration, right?

vazmin: Yes, `build.options.extractCss ` is already true, ` build.configurations.production.extractCss` can be removed.

PR #750, opened 2020-04-22
auto load examples from @angular/components-examples

annieyw:


src/app/shared/documentation-items/documentation-items.ts
```diff
@@ -66,111 +75,78 @@ const DOCS: {[key: string]: DocCategory[]} = {
...
+          },
         },
         {
           id: 'radio',
           name: 'Radio button',
           summary: 'Allows the user to select one option from a group.',
-          examples: ['radio-ng-model'],
+          exampleSpecs: {
+            prefix: 'radio-',
+            exclude: [],
```
atscott, line 122: If all the `exclude` arrays are empty, should they just be removed?

src/app/shared/documentation-items/documentation-items.ts
```diff
@@ -361,37 +340,35 @@ const DOCS: {[key: string]: DocCategory[]} = {
           id: 'bottom-sheet',
           name: 'Bottom Sheet',
           summary: 'A large interactive panel primarily for mobile devices.',
-          examples: ['bottom-sheet-overview']
+          exampleSpecs: {
+            prefix: 'bottom-sheet',
```
atscott, line 344: Curious why some prefixes end with a `-` but others don't. Is this intended?

src/app/shared/documentation-items/documentation-items.ts
```diff
@@ -449,88 +415,102 @@ const DOCS: {[key: string]: DocCategory[]} = {
           id: 'a11y',
           name: 'Accessibility',
           summary: 'Utilities for screen readers, focus and more.',
-          examples: []
+          exampleSpecs: {
+            prefix: 'focus-monitor',
+            exclude: [],
+          },
+          // examples: []
```
atscott, line 422: nit: remove

src/app/shared/documentation-items/documentation-items.ts
annieyw: Oops it was a mistake. Now they all end with `-`

src/app/shared/documentation-items/documentation-items.ts
```diff
@@ -601,12 +570,20 @@ const DOCS: {[key: string]: DocCategory[]} = {
 for (const category of DOCS[COMPONENTS]) {
   for (const doc of category.items) {
     doc.packageName = 'material';
+    doc.examples =
+      exampleNames
+        .filter(key => key.match(doc.exampleSpecs.prefix) &&
```
mmalerba, line 575: probably should be `.startsWith` to make sure we really are treating it as a prefix

annieyw: `.startsWith` not support on IE :( 
added `^` to regex to match the start.

src/app/shared/documentation-items/documentation-items.ts
```diff
@@ -601,12 +570,20 @@ const DOCS: {[key: string]: DocCategory[]} = {
 for (const category of DOCS[COMPONENTS]) {
   for (const doc of category.items) {
     doc.packageName = 'material';
+    doc.examples =
+      exampleNames
+        .filter(key => key.match(/^${doc.exampleSpecs.prefix}/) &&
```
mmalerba, line 575: I don't think this syntax works in regexes, you'll need to do ``RegExp(`^${...}`)``

PR #751, opened 2020-04-23
refactor: create a service to manage navigation & focus state in the docs app

annieyw:
Currently there are several places in the app try to track navigation events and properly redirect focus. The current approach is rather brittle, it would be better to coordinate it through a globally available service.

This service can manage:
- Allowing directives to request that the skip link jump to them
- Deciding which element will actually receive focus from the skip link
- Differentiating between hard and soft navigations

### DIFF
--- a/src/app/pages/component-viewer/component-viewer.html
+++ b/src/app/pages/component-viewer/component-viewer.html
@@ -1,7 +1,7 @@
 <div class="docs-component-viewer">
   <nav mat-tab-nav-bar class="docs-component-viewer-tabbed-content"
        aria-label="Documentation Sections"
-       focusOnNavigation id="main-content" >
+       focusOnNavigation >
     <a mat-tab-link class="docs-component-viewer-section-tab"
         *ngFor="let section of sections"
         [routerLink]="section.toLowerCase()"
--- a/src/app/pages/component-viewer/component-viewer.html
+++ b/src/app/pages/component-viewer/component-viewer.html
@@ -1,7 +1,7 @@
 <div class="docs-component-viewer">
   <nav mat-tab-nav-bar class="docs-component-viewer-tabbed-content"
        aria-label="Documentation Sections"
-       id = "component-viewer"
+       id="component-viewer"
        focusOnNavigation >
     <a mat-tab-link class="docs-component-viewer-section-tab"
         *ngFor="let section of sections"
--- a/src/app/pages/component-viewer/component-viewer.html
+++ b/src/app/pages/component-viewer/component-viewer.html
@@ -1,6 +1,7 @@
 <div class="docs-component-viewer">
   <nav mat-tab-nav-bar class="docs-component-viewer-tabbed-content"
        aria-label="Documentation Sections"
+       id = "component-viewer"
        focusOnNavigation >
     <a mat-tab-link class="docs-component-viewer-section-tab"
         *ngFor="let section of sections"
--- a/src/app/shared/navigation-focus/navigation-focus.service.ts
+++ b/src/app/shared/navigation-focus/navigation-focus.service.ts
@@ -43,7 +43,7 @@ export class NavigationFocusService implements OnDestroy {
   }
 
   requestSkipLinkFocus(el: ElementRef) {
-    this.navigationFocusRequests.push(el);
+    this.skipLinkFocusRequests.push(el);
     const baseUrl = this.router.url.split('#')[0];
     const skipLinKTargetId = el.nativeElement.id;
     this.skipLinkHref = `${baseUrl}#${skipLinKTargetId}`;
@@ -51,6 +51,12 @@ export class NavigationFocusService implements OnDestroy {
 
   relinquishSkipLinkFocusOnDestroy(el: ElementRef) {
     this.skipLinkFocusRequests.splice(this.skipLinkFocusRequests.indexOf(el), 1);
+    if (this.skipLinkFocusRequests.length) {
+      const skipLinkFocusTarget = this.skipLinkFocusRequests[this.skipLinkFocusRequests.length - 1];
+      const baseUrl = this.router.url.split('#')[0];
+      const skipLinKTargetId = skipLinkFocusTarget.nativeElement.id;
+      this.skipLinkHref = `${baseUrl}#${skipLinKTargetId}`;
+    }
   }
 
   getSkipLinkHref(): string {
--- a/src/app/shared/navigation-focus/navigation-focus.service.ts
+++ b/src/app/shared/navigation-focus/navigation-focus.service.ts
@@ -34,20 +34,23 @@ export class NavigationFocusService implements OnDestroy {
     this.subscriptions.unsubscribe();
   }
 
-  requestFocusOnNavigation(el: ElementRef, wantsFocus: boolean) {
-    wantsFocus ? this.navigationFocusRequests.push(el) :
-      this.navigationFocusRequests.splice(this.navigationFocusRequests.indexOf(el), 1);
+  requestFocusOnNavigation(el: ElementRef) {
+    this.navigationFocusRequests.push(el);
   }
 
-  requestSkipLinkFocus(el: ElementRef, wantsFocus: boolean) {
-    if (wantsFocus) {
-      this.navigationFocusRequests.push(el);
-      const baseUrl = this.router.url.split('#')[0];
-      const skipLinKTargetId = el.nativeElement.id;
-      this.skipLinkHref = `${baseUrl}#${skipLinKTargetId}`;
-    } else {
-      this.skipLinkFocusRequests.splice(this.skipLinkFocusRequests.indexOf(el), 1);
-    }
+  relinquishFocusOnDestroy(el: ElementRef) {
+    this.navigationFocusRequests.splice(this.navigationFocusRequests.indexOf(el), 1);
+  }
+
+  requestSkipLinkFocus(el: ElementRef) {
+    this.navigationFocusRequests.push(el);
+    const baseUrl = this.router.url.split('#')[0];
+    const skipLinKTargetId = el.nativeElement.id;
+    this.skipLinkHref = `${baseUrl}#${skipLinKTargetId}`;
+  }
+
+  relinquishSkipLinkFocusOnDestroy(el: ElementRef) {
+    this.skipLinkFocusRequests.splice(this.skipLinkFocusRequests.indexOf(el), 1);
   }
 
   getSkipLinkHref(): string {
--- a/src/app/shared/navigation-focus/navigation-focus.service.ts
+++ b/src/app/shared/navigation-focus/navigation-focus.service.ts
@@ -1,0 +1,31 @@
+import {ElementRef, Injectable} from '@angular/core';
+import {Event, NavigationEnd, Router} from '@angular/router';
+import {filter, skip, take} from 'rxjs/operators';
+
+@Injectable({
+  providedIn: 'root',
+})
+export class NavigationFocusService {
+  private internalNavigation: boolean;
+  private focusRequestStack: ElementRef[];
+
+  constructor(private router:Router) {
+    this.router.events
+      .pipe(filter((event: Event): event is NavigationEnd => event instanceof NavigationEnd),
+        skip(1),
+        take(1)
+      )
+      .subscribe(() => {
+        this.internalNavigation = true;
+      });
+  }
+
+  isInternalNavigation(): boolean {
+    return this.internalNavigation;
+  }
+
+  requestFocusOnNavigation(el: ElementRef, wantsFocus: boolean) {
+    wantsFocus ? this.focusRequestStack.push(el) :
+      this.focusRequestStack.splice(this.focusRequestStack.indexOf(el), 1);
+  }
+}
--- a/src/app/shared/navigation-focus/navigation-focus.service.ts
+++ b/src/app/shared/navigation-focus/navigation-focus.service.ts
@@ -1,34 +1,29 @@
 import {ElementRef, Injectable} from '@angular/core';
 import {NavigationEnd, Router} from '@angular/router';
-import {filter, skip, take} from 'rxjs/operators';
+import {filter, skip} from 'rxjs/operators';
 
 @Injectable({
   providedIn: 'root',
 })
 export class NavigationFocusService {
   private navigationFocusRequests: ElementRef[] = [];
-  private skipLinkFocusRequests: ElementRef[] = [];
+  private skipLinkHref = '';
+
+  readonly navigationEndEvents = this.router.events.pipe(filter(event => event instanceof
+    NavigationEnd));
+  readonly softNavigations = this.navigationEndEvents.pipe(skip(1));
 
   constructor(private router: Router) {
-    this.router.events
-      .pipe(filter(event => event instanceof NavigationEnd),
-        take(1))
-      .subscribe(() => {
-        setTimeout(() =>
-          this.skipLinkFocusRequests[this.skipLinkFocusRequests.length - 1].nativeElement.id =
-            'main-content', 100);
-      });
-
-    this.router.events
-      .pipe(filter(event => event instanceof NavigationEnd),
-        skip(1))
-      .subscribe(() => {
-        if (!window.location.hash) {
-          setTimeout(() =>
+    this.softNavigations.subscribe(() => {
+      // focus if url does not have fragment
+      if (!this.router.url.split('#')[1]) {
+        setTimeout(() => {
+          if (this.navigationFocusRequests.length) {
             this.navigationFocusRequests[this.navigationFocusRequests.length - 1]
-              .nativeElement.focus({preventScroll: true}), 100);
-        }
-      });
+              .nativeElement.focus({preventScroll: true});
+          }}, 100);
+      }
+    });
   }
 
   requestFocusOnNavigation(el: ElementRef, wantsFocus: boolean) {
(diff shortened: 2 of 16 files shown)


src/app/shared/navigation-focus/navigation-focus.service.ts
```diff
@@ -0,0 +1,25 @@
+import {Injectable} from '@angular/core';
+import {Event, NavigationEnd, Router} from '@angular/router';
+import {filter} from 'rxjs/operators';
+
+@Injectable()
```
mmalerba, line 5: I would probably make this a singleton service: https://angular.io/guide/singleton-services#providing-a-singleton-service

src/app/shared/navigation-focus/navigation-focus.service.ts
```diff
@@ -0,0 +1,43 @@
...
+  private navigationFocusRequests: ElementRef[] = [];
+  private skipLinkFocusRequests: ElementRef[] = [];
+
+  constructor(private router: Router) {
+    this.router.events
+      .pipe(filter(event => event instanceof NavigationEnd),
+        take(1))
+      .subscribe(() => {
+        setTimeout(() =>
+          this.skipLinkFocusRequests[this.skipLinkFocusRequests.length - 1].nativeElement.id =
```
mmalerba, line 18: I don't think we need to wait for a navigation request to do this, we can just do it right when they request.

Also, I would prefer if we have an id on the element and then update the skiplink to refer to that ID, rather than changing the ID on the element

src/app/shared/navigation-focus/navigation-focus.service.ts
```diff
@@ -0,0 +1,43 @@
...
+            'main-content', 100);
+      });
+
+    this.router.events
+      .pipe(filter(event => event instanceof NavigationEnd),
+        skip(1))
+      .subscribe(() => {
+        if (!window.location.hash) {
```
mmalerba, line 26: is there still code somewhere that handles the fragment links or did we remove that?

annieyw: I think that was removed

src/app/shared/navigation-focus/navigation-focus.service.ts
```diff
@@ -0,0 +1,43 @@
...
+@Injectable({
+  providedIn: 'root',
+})
+export class NavigationFocusService {
+  private navigationFocusRequests: ElementRef[] = [];
+  private skipLinkFocusRequests: ElementRef[] = [];
+
+  constructor(private router: Router) {
+    this.router.events
+      .pipe(filter(event => event instanceof NavigationEnd),
```
atscott, line 14: Would it be helpful to make this a generic `NavigationUtils` service and publish this pipe like
```
readonly navigationEndEvents = this.router.events.pipe(filter(event => event instanceof NavigationEnd));
readonly softNavigations = this.navigationEndEvents.pipe(skip(1));
```
There are several places in the project that could use this.

src/app/shared/navigation-focus/navigation-focus.service.ts
atscott: Can you get the fragment from the `Router` state instead or at least from some injectable service? Otherwise you'll have trouble testing this.

src/app/shared/navigation-focus/navigation-focus.service.ts
```diff
@@ -0,0 +1,43 @@
+import {ElementRef, Injectable} from '@angular/core';
+import {NavigationEnd, Router} from '@angular/router';
+import {filter, skip, take} from 'rxjs/operators';
+
+@Injectable({
+  providedIn: 'root',
+})
+export class NavigationFocusService {
```
atscott, line 8: Please write some tests for this service.

src/app/shared/navigation-focus/navigation-focus.service.ts
mmalerba: I think the hope is to pull this all of the router subscriptions under the scope of this service anyways

mmalerba: Oh nevermind, I see there are a few places that use it for things other than redirecting focus

src/app/shared/navigation-focus/navigation-focus.service.ts
```diff
@@ -0,0 +1,43 @@
...
+      }
+    });
+  }
+
+  requestFocusOnNavigation(el: ElementRef, wantsFocus: boolean) {
+    wantsFocus ? this.navigationFocusRequests.push(el) :
+      this.navigationFocusRequests.splice(this.navigationFocusRequests.indexOf(el), 1);
+  }
+
+  requestSkipLinkFocus(el: ElementRef) {
```
mmalerba, line 34: I think we want to maintain a list for this too, so we know where to send the skiplink when a component unloads

src/app/shared/navigation-focus/navigation-focus.ts
```diff
@@ -1,49 +1,23 @@
...
-
-/** The timeout id of the previous focus change. */
-let lastTimeoutId = -1;
+import {NavigationFocusService} from './navigation-focus.service';
 
 @Directive({
   selector: '[focusOnNavigation]',
 })
 export class NavigationFocus implements OnDestroy {
   @HostBinding('tabindex') role = '-1';
```
mmalerba, line 8: I haven no idea why the property is called `role`, can you change it to `tabindex`

src/app/shared/navigation-focus/navigation-focus.service.ts
```diff
@@ -0,0 +1,43 @@
...
+export class NavigationFocusService {
+  private navigationFocusRequests: ElementRef[] = [];
+  private skipLinkHref = '';
+
+  readonly navigationEndEvents = this.router.events.pipe(filter(event => event instanceof
+    NavigationEnd));
+  readonly softNavigations = this.navigationEndEvents.pipe(skip(1));
+
+  constructor(private router: Router) {
+    this.softNavigations.subscribe(() => {
```
mmalerba, line 17: unsubscribe

src/app/shared/navigation-focus/navigation-focus.service.ts
```diff
@@ -0,0 +1,67 @@
...
+  ngOnDestroy() {
+    this.subscriptions.unsubscribe();
+  }
+
+  requestFocusOnNavigation(el: ElementRef, wantsFocus: boolean) {
+    wantsFocus ? this.navigationFocusRequests.push(el) :
+      this.navigationFocusRequests.splice(this.navigationFocusRequests.indexOf(el), 1);
+  }
+
+  requestSkipLinkFocus(el: ElementRef, wantsFocus: boolean) {
```
atscott, line 42: Can we refactor these two be different functions (2 instead of 4) without the boolean? The function bodies have completely different behaviors based on the boolean value and don't share any code. I feel like this is the purest example of the boolean argument as a "code smell".

src/app/shared/navigation-focus/navigation-focus.spec.ts
```diff
@@ -0,0 +1,33 @@
...
+    navigationFocusService = nfs;
+  }));
+
+  it('should set skip link href', () => {
+    const div = document.createElement('div');
+    div.id = 'skip-link-target';
+    const element = new ElementRef(div);
+    navigationFocusService.requestSkipLinkFocus(element, true);
+    expect(navigationFocusService.getSkipLinkHref()).toBe('/#skip-link-target');
+  });
```
atscott, line 32: Needs more test coverage: `isNavigationWithinComponentView` is untested, `requestSkipLinkFocus` with `false` is untested, `requestFocusOnNavigation` is untested.

src/app/shared/navigation-focus/navigation-focus.ts
```diff
@@ -1,47 +1,26 @@
...
 
 @Directive({
   selector: '[focusOnNavigation]',
 })
 export class NavigationFocus implements OnDestroy {
-  @HostBinding('tabindex') role = '-1';
-
-  private subscriptions = new Subscription();
+  @HostBinding('tabindex') tabindex = '-1';
+  @HostBinding('style.outline') outline = 'none';
```
atscott, line 9: Should these be `readonly`?

src/app/shared/navigation-focus/navigation-focus.spec.ts
```diff
@@ -0,0 +1,103 @@
...
+          {path: 'guides', component: RouteTest}
+        ])],
+        providers: [NavigationFocusService],
+        declarations: [NavigationFocusTest, RouteTest],
+      });
+      fixture = TestBed.createComponent(NavigationFocusTest);
+    }
+  );
+
+  beforeEach(inject(
```
atscott, line 32: Can you try updating this to use `TestBed.inject` instead of `beforeEach(inject(...))`?
i.e.
```
beforeEach(() => {
  zone = TestBed.inject(NgZone);
etc
```

src/app/shared/navigation-focus/navigation-focus.spec.ts
```diff
@@ -0,0 +1,103 @@
...
+    expect(navigationFocusService.isNavigationWithinComponentView(previousUrl, newUrl)).toBeTrue();
+  });
+
+  it('should not be within component view', () => {
+    const previousUrl = '/cdk/clipboard/overview';
+    const newUrl = '/cdk/categories';
+    expect(navigationFocusService.isNavigationWithinComponentView(previousUrl, newUrl)).toBeFalse();
+  });
+
+  it('should focus on component then relinquish focus', fakeAsync(() => {
```
atscott, line 66: Excellent test. It feels really satisfying to see tests demonstrate expected behavior so clearly. Can you also add one that exercises the path where you navigate to a hash target within the page? And one that ensures focus is still applied to elements that didn't have an id (the path where it adds one using the uid).

src/app/shared/navigation-focus/navigation-focus.spec.ts
mmalerba: Is this the new recommended way to do it? I've always done the `beforeEach(inject(...))` style

src/app/shared/navigation-focus/navigation-focus.service.ts
```diff
@@ -0,0 +1,76 @@
...
+
+  ngOnDestroy() {
+    this.subscriptions.unsubscribe();
+  }
+
+  requestFocusOnNavigation(el: HTMLElement) {
+    this.navigationFocusRequests.push(el);
+  }
+
+  relinquishFocusOnDestroy(el: HTMLElement) {
```
mmalerba, line 41: nit: can we call this `relinquishFocusOnNavigation`

src/app/shared/navigation-focus/navigation-focus.service.ts
```diff
@@ -0,0 +1,76 @@
...
+  }
+
+  requestSkipLinkFocus(el: HTMLElement) {
+    this.skipLinkFocusRequests.push(el);
+    const baseUrl = this.router.url.split('#')[0];
+    const skipLinKTargetId = el.id;
+    this.skipLinkHref = `${baseUrl}#${skipLinKTargetId}`;
+  }
+
+  relinquishSkipLinkFocusOnDestroy(el: HTMLElement) {
```
mmalerba, line 52: nit: `relinquishSkipLinkFocus`

src/app/shared/navigation-focus/navigation-focus.service.ts
```diff
@@ -0,0 +1,76 @@
...
+    this.navigationFocusRequests.push(el);
+  }
+
+  relinquishFocusOnDestroy(el: HTMLElement) {
+    this.navigationFocusRequests.splice(this.navigationFocusRequests.indexOf(el), 1);
+  }
+
+  requestSkipLinkFocus(el: HTMLElement) {
+    this.skipLinkFocusRequests.push(el);
+    const baseUrl = this.router.url.split('#')[0];
```
mmalerba, line 47: this code is similar to the code in the relinquish method, it can probably be extracted into a helper

src/app/shared/navigation-focus/navigation-focus.service.ts
```diff
@@ -0,0 +1,76 @@
...
+    const skipLinKTargetId = el.id;
+    this.skipLinkHref = `${baseUrl}#${skipLinKTargetId}`;
+  }
+
+  relinquishSkipLinkFocusOnDestroy(el: HTMLElement) {
+    this.skipLinkFocusRequests.splice(this.skipLinkFocusRequests.indexOf(el), 1);
+    if (this.skipLinkFocusRequests.length) {
```
mmalerba, line 54: this isn't quite right because it doesn't set the href to ''/null when everything is popped

src/app/shared/navigation-focus/navigation-focus.spec.ts
atscott: I would say so, yea. If I recall correctly `beforeEach(inject(...))` was the way to do it before `TestBed.inject` existed. The testing guide https://angular.io/guide/testing uses `TestBed.inject` and the angular cli does as well when you generate a new service.

src/app/pages/component-viewer/component-viewer.html
```diff
@@ -1,7 +1,8 @@
 <div class="docs-component-viewer">
   <nav mat-tab-nav-bar class="docs-component-viewer-tabbed-content"
        aria-label="Documentation Sections"
-       focusOnNavigation id="main-content" >
+       id = "component-viewer"
```
mmalerba, line 4: nit: no spaces around `=`

src/app/shared/navigation-focus/navigation-focus.service.ts
```diff
@@ -0,0 +1,84 @@
...
+      this.setSkipLinkHref(skipLinkFocusTarget);
+    } else {
+      this.skipLinkHref = '';
+    }
+  }
+
+  setSkipLinkHref(el: HTMLElement) {
+    const baseUrl = this.router.url.split('#')[0];
+    if (!el.id) {
+      el.id = `skip-link-target-${this.uid++}`;
```
mmalerba, line 64: can you move this logic to the directive? that seems like a more natural place to have it

src/app/shared/navigation-focus/navigation-focus.service.ts
```diff
@@ -0,0 +1,84 @@
...
+    this.skipLinkFocusRequests.splice(this.skipLinkFocusRequests.indexOf(el), 1);
+    if (this.skipLinkFocusRequests.length) {
+      const skipLinkFocusTarget = this.skipLinkFocusRequests[this.skipLinkFocusRequests.length - 1];
+      this.setSkipLinkHref(skipLinkFocusTarget);
+    } else {
+      this.skipLinkHref = '';
+    }
+  }
+
+  setSkipLinkHref(el: HTMLElement) {
```
mmalerba, line 61: if you make this take `HTMLElement|null` you don't need the if logic in `relinquishSkipLinkFocus`

src/app/shared/table-of-contents/table-of-contents.ts
```diff
@@ -47,16 +48,16 @@ export class TableOfContents implements OnInit, AfterViewInit, OnDestroy {
   constructor(private _router: Router,
               private _route: ActivatedRoute,
               private _element: ElementRef,
+              private _navigationFocusService: NavigationFocusService,
               @Inject(DOCUMENT) private _document: Document) {
 
-    this._router.events.pipe(takeUntil(this._destroyed)).subscribe((event) => {
-      if (event instanceof NavigationEnd) {
+    this._navigationFocusService.navigationEndEvents.pipe(takeUntil(this._destroyed))
```
mmalerba, line 54: nit: can you switch this to use the `subscriptions.add` style unsubscribe

src/app/shared/navigation-focus/navigation-focus.service.ts
```diff
@@ -0,0 +1,75 @@
...
+
+  relinquishSkipLinkFocus(el: HTMLElement) {
+    this.skipLinkFocusRequests.splice(this.skipLinkFocusRequests.indexOf(el), 1);
+    const skipLinkFocusTarget = this.skipLinkFocusRequests[this.skipLinkFocusRequests.length - 1];
+    this.setSkipLinkHref(skipLinkFocusTarget);
+  }
+
+  setSkipLinkHref(el: HTMLElement|null) {
+    const baseUrl = this.router.url.split('#')[0];
+    this.skipLinkHref = el ? `${baseUrl}#${el.id}` : '/null';
```
mmalerba, line 58: I think we just want `null`, not the string `'null'`

src/app/shared/navigation-focus/navigation-focus.spec.ts
```diff
@@ -0,0 +1,128 @@
...
+
+    expect(navigationFocusService.getSkipLinkHref()).toEqual('/#target1');
+  });
+
+  it('should set id for skip link target without id', () => {
+    const skipLinkTarget = fixture.nativeElement.querySelector('.no-id');
+
+    navigationFocusService.requestSkipLinkFocus(skipLinkTarget);
+
+    expect(navigationFocusService.getSkipLinkHref()).toMatch(`\#skip-link-target-[0-9]*`);
```
mmalerba, line 59: I don't think you have to escape `#` in a string, or was that supposed to be a forward slash?

src/app/shared/navigation-focus/navigation-focus.spec.ts
```diff
@@ -0,0 +1,128 @@
...
+    navigationFocusService.requestSkipLinkFocus(target2);
+
+    expect(navigationFocusService.getSkipLinkHref()).toEqual('/#target2');
+
+    navigationFocusService.relinquishSkipLinkFocus(target2);
+
+    expect(navigationFocusService.getSkipLinkHref()).toEqual('/#target1');
+  });
+
+  it('should set id for skip link target without id', () => {
```
mmalerba, line 54: can you add a test that the href is null when everything relinquishes?

mmalerba:
Thanks for making these changes! This should make it a lot easier to maintain the navigation logic going forward

PR #758, opened 2020-05-08
enable compact view examples

annieyw:
Extract fields from html generated from embedded example comments in md docs to generate compact view examples.

(extra: update button colours to be consistent)

### DIFF
--- a/src/app/shared/doc-viewer/doc-viewer.ts
+++ b/src/app/shared/doc-viewer/doc-viewer.ts
@@ -113,6 +113,8 @@ export class DocViewer implements OnDestroy {
 
     [...exampleElements].forEach((element: Element) => {
       const example = element.getAttribute(componentName);
+      const lines = element.getAttribute('lines');
+      const file = element.getAttribute('file');
       const portalHost = new DomPortalOutlet(
           element, this._componentFactoryResolver, this._appRef, this._injector);
       const examplePortal = new ComponentPortal(componentClass, this._viewContainerRef);
@@ -120,7 +122,14 @@ export class DocViewer implements OnDestroy {
       if (example !== null) {
         const exampleViewerComponent = exampleViewer.instance as ExampleViewer;
         exampleViewerComponent.example = example;
-        exampleViewerComponent.view = 'collapsed';
+        if (file) {
+          exampleViewerComponent.view = 'compact';
+          exampleViewerComponent.showCompactToggle = true;
+          exampleViewerComponent.file = file;
+        }
+        if (lines) {
+          exampleViewerComponent.lines = JSON.parse(lines);
+        }
       }
 
       this._portalHosts.push(portalHost);
--- a/src/app/shared/doc-viewer/doc-viewer.ts
+++ b/src/app/shared/doc-viewer/doc-viewer.ts
@@ -38,12 +38,37 @@ export class DocViewer implements OnDestroy {
     }
   }
 
-  @Input() lines: readonly [number, number];
-
   @Output() contentRendered = new EventEmitter<HTMLElement>();
 
   /** The document text. It should not be HTML encoded. */
   textContent = '';
+
+  private static initExampleViewer(exampleViewerComponent: ExampleViewer,
+                                   example: string,
+                                   expanded: string | null,
+                                   file: string | null,
+                                   region: string | null) {
+    exampleViewerComponent.example = example;
+    if (expanded === 'true') {
+      exampleViewerComponent.view = 'full';
+    } else {
+      if (file) {
+        // if the html div has field `file` then it should be in compact view to show the code
+        // snippet
+        exampleViewerComponent.view = 'snippet';
+        exampleViewerComponent.showCompactToggle = true;
+        exampleViewerComponent.file = file;
+        if (region) {
+          // `lines` should only exist when `file` exists but not vice versa
+          // it is valid for embedded example snippets to show the whole file (esp short files)
+          exampleViewerComponent.region = region;
+        }
+      } else {
+        // otherwise it is an embedded demo
+        exampleViewerComponent.view = 'demo';
+      }
+    }
+  }
 
   constructor(private _appRef: ApplicationRef,
               private _componentFactoryResolver: ComponentFactoryResolver,
@@ -82,12 +107,6 @@ export class DocViewer implements OnDestroy {
     });
     this._elementRef.nativeElement.innerHTML = rawDocument;
     this.textContent = this._elementRef.nativeElement.textContent;
-    if (this.lines) {
-      this._elementRef.nativeElement.innerHTML =
-        rawDocument.split('\n').slice(this.lines[0], this.lines[1]).join('\n');
-      this.textContent = this._elementRef.nativeElement.textContent;
-    }
-
     this._loadComponents('material-docs-example', ExampleViewer);
     this._loadComponents('header-link', HeaderLink);
 
@@ -113,25 +132,17 @@ export class DocViewer implements OnDestroy {
 
     [...exampleElements].forEach((element: Element) => {
       const example = element.getAttribute(componentName);
-      const lines = element.getAttribute('lines');
+      const region = element.getAttribute('region');
       const file = element.getAttribute('file');
+      const expanded = element.getAttribute('expanded');
       const portalHost = new DomPortalOutlet(
           element, this._componentFactoryResolver, this._appRef, this._injector);
       const examplePortal = new ComponentPortal(componentClass, this._viewContainerRef);
       const exampleViewer = portalHost.attach(examplePortal);
+      const exampleViewerComponent = exampleViewer.instance as ExampleViewer;
       if (example !== null) {
-        const exampleViewerComponent = exampleViewer.instance as ExampleViewer;
-        exampleViewerComponent.example = example;
-        if (file) {
-          exampleViewerComponent.view = 'compact';
-          exampleViewerComponent.showCompactToggle = true;
-          exampleViewerComponent.file = file;
-        }
-        if (lines) {
-          exampleViewerComponent.lines = JSON.parse(lines);
-        }
+        DocViewer.initExampleViewer(exampleViewerComponent, example, expanded, file, region);
       }
-
       this._portalHosts.push(portalHost);
     });
   }
--- a/src/app/shared/doc-viewer/doc-viewer.ts
+++ b/src/app/shared/doc-viewer/doc-viewer.ts
@@ -45,29 +45,25 @@ export class DocViewer implements OnDestroy {
 
   private static initExampleViewer(exampleViewerComponent: ExampleViewer,
                                    example: string,
-                                   expanded: string | null,
                                    file: string | null,
                                    region: string | null) {
     exampleViewerComponent.example = example;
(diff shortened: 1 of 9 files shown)


src/app/shared/doc-viewer/doc-viewer.ts
```diff
@@ -113,14 +113,23 @@ export class DocViewer implements OnDestroy {
...
       const portalHost = new DomPortalOutlet(
           element, this._componentFactoryResolver, this._appRef, this._injector);
       const examplePortal = new ComponentPortal(componentClass, this._viewContainerRef);
       const exampleViewer = portalHost.attach(examplePortal);
       if (example !== null) {
         const exampleViewerComponent = exampleViewer.instance as ExampleViewer;
         exampleViewerComponent.example = example;
-        exampleViewerComponent.view = 'collapsed';
+        if (file) {
+          exampleViewerComponent.view = 'compact';
```
mmalerba, line 126: Isn't there some boolean attribute that tells us whether it should be compact or full?

src/app/shared/example-viewer/example-viewer.html
```diff
@@ -52,10 +52,10 @@
         </mat-icon>
       </button>
     </div>
-    <code-snippet [source]="file" [lines]="lines" *ngIf="file !== 'demo'"></code-snippet>
+    <code-snippet [source]="generateUrl(file)" [lines]="lines"></code-snippet>
```
mmalerba, line 55: Doing it this way means we have to redo the work to generate the URL on every change detection. Instead you can make the file input a getter/setter and do the URL generation in the setter.

i.e. in the TS:

```ts
@Input()
get file() { return this._file; }
set file(relativeFile) {
  this._file = generateUrl(relativeFile);
}
```

and then in the HTML just:
```html
<... [source]="file" ...>
```

annieyw: I had to put it in `ngOnInit` otherwise `this.example` and `this.exampleData.module.importSpecifier` would be missing

src/app/shared/doc-viewer/doc-viewer.ts
```diff
@@ -113,14 +113,35 @@ export class DocViewer implements OnDestroy {
...
+      const expanded = element.getAttribute('expanded');
       const portalHost = new DomPortalOutlet(
           element, this._componentFactoryResolver, this._appRef, this._injector);
       const examplePortal = new ComponentPortal(componentClass, this._viewContainerRef);
       const exampleViewer = portalHost.attach(examplePortal);
       if (example !== null) {
         const exampleViewerComponent = exampleViewer.instance as ExampleViewer;
         exampleViewerComponent.example = example;
-        exampleViewerComponent.view = 'collapsed';
+        if (expanded === 'true') {
```
mmalerba, line 126: let's move the stuff inside `if(example !== null)` into a new method, e.g. `initExampleViewer`, since this method is getting kinda big

src/app/shared/example-viewer/example-viewer.ts
```diff
@@ -67,6 +67,12 @@ export class ExampleViewer {
   constructor(private readonly snackbar: MatSnackBar, private readonly copier: CopierService) {
   }
 
+  ngOnInit() {
+    if (this.file) {
+      this.file = this.generateUrl(this.file);
```
mmalerba, line 72: I don't really like this since `file` has different semantics before and after `ngOnInit` runs. I would just save it in a different property like `fileUrl`

mmalerba:
Whoops, looks like CI is broken, good to go once its fixed though

src/app/shared/example-viewer/example-viewer.ts
```diff
@@ -59,20 +65,42 @@ export class ExampleViewer {
...
   }
 
+  ngOnInit() {
+    if (this.file) {
+      this.fileUrl = this.generateUrl(this.file);
+    }
+  }
+
+  selectCorrectTab() {
+    const fileType = this.file.split('.')[1];
```
mmalerba, line 83: you should grab the last chunk, not the second - this could be different if the file is named something like `stuff.spec.ts`

mmalerba: also add a test to cover this

src/app/shared/example-viewer/example-viewer.ts
```diff
@@ -83,6 +111,19 @@ export class ExampleViewer {
     }
   }
 
+  generateUrl(file: string): string {
+    let fileName: string;
+    if (this.region !== 'undefined') {
```
mmalerba, line 116: the string `'undefined'`?

src/app/shared/example-viewer/example-viewer.ts
```diff
@@ -83,6 +111,19 @@ export class ExampleViewer {
...
+      const fileSplit = file.split('.');
+      fileName = fileSplit[0] + '_' + this.region + '-' + fileSplit[1] + '.html';
```
mmalerba, line 118: again, consider files that have multiple `.` just to be safe

src/app/shared/example-viewer/example-viewer.ts
atscott: I'd also consider throwing an error/logging a warning if the file type is something other than the ones listed.

src/app/shared/example-viewer/code-snippet.html
```diff
@@ -1,3 +1,3 @@
 <div class="docs-example-source-wrapper">
-  <pre class="docs-example-source"><doc-viewer #viewer [documentUrl]="source" [lines]="lines"></doc-viewer></pre>
+  <pre class="docs-example-source"><doc-viewer #viewer [documentUrl]="source" ></doc-viewer></pre>
```
atscott, line 2: nit: extra space

src/app/shared/doc-viewer/doc-viewer.ts
```diff
@@ -38,13 +38,38 @@ export class DocViewer implements OnDestroy {
...
+        exampleViewerComponent.file = file;
+        if (region) {
+          // `lines` should only exist when `file` exists but not vice versa
+          // it is valid for embedded example snippets to show the whole file (esp short files)
+          exampleViewerComponent.region = region;
+        }
+      } else {
+        // otherwise it is an embedded demo
+        exampleViewerComponent.view = 'demo';
+      }
```
atscott, line 69: Can we extract something in here to a helper method? It's visually hard to keep track of the `if...else` nesting.

src/app/shared/example-viewer/example-viewer.ts
```diff
@@ -59,20 +65,42 @@ export class ExampleViewer {
   private _example: string;
 
   /** Range of lines of the source code to display in compact view. */
-  @Input() lines: readonly [number, number];
+  @Input() region: string;
```
atscott, line 68: I think this should be optional.
```suggestion
  @Input() region?: string;
```

src/app/shared/example-viewer/example-viewer.ts
```diff
@@ -59,20 +65,42 @@ export class ExampleViewer {
...
 
   /** Name of file to display in compact view. */
   @Input() file: string;
```
atscott, line 71: Should this also be optional? Based on the code in `ngOnInit`, it seems so.

src/app/shared/doc-viewer/doc-viewer.ts
```diff
@@ -38,13 +38,38 @@ export class DocViewer implements OnDestroy {
...
+    } else {
+      if (file) {
+        // if the html div has field `file` then it should be in compact view to show the code
+        // snippet
+        exampleViewerComponent.view = 'snippet';
+        exampleViewerComponent.showCompactToggle = true;
+        exampleViewerComponent.file = file;
+        if (region) {
+          // `lines` should only exist when `file` exists but not vice versa
+          // it is valid for embedded example snippets to show the whole file (esp short files)
```
atscott, line 63:
```suggestion
          // `region` should only exist when `file` exists but not vice versa.
          // It is valid for embedded example snippets to show the whole file (esp short files).
```

src/app/shared/doc-viewer/doc-viewer.ts
```diff
@@ -38,13 +38,38 @@ export class DocViewer implements OnDestroy {
...
-  @Input() lines: readonly [number, number];
-
   @Output() contentRendered = new EventEmitter<HTMLElement>();
 
   /** The document text. It should not be HTML encoded. */
   textContent = '';
 
+  private static initExampleViewer(exampleViewerComponent: ExampleViewer,
+                                   example: string,
+                                   expanded: string | null,
```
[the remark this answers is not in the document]
annieyw: decided to remove `expanded` from API because it is no longer necessary after the other API changes made

src/app/shared/doc-viewer/doc-viewer.ts
annieyw: after removing `expanded` there is 1 less level, do you think that is sufficiently simple or should some parts still be extracted? I

atscott: I think this is good now 👍

src/app/shared/example-viewer/example-viewer.spec.ts
```diff
@@ -48,12 +48,57 @@ describe('ExampleViewer', () => {
...
+  }));
+
+  it('should print an error message about incorrect file type', async(() => {
+    spyOn(console, 'error');
+    fixture.detectChanges();
+    component.file = 'file.bad';
+    component.selectCorrectTab();
+
+
+    expect(console.error).toHaveBeenCalled();
```
atscott, line 99: nit: you can remove this. It's covered by the line below. Error would be 
`expected console.error to have been called with xxxx, but was never called`

src/app/shared/example-viewer/example-viewer.spec.ts
```diff
@@ -48,12 +48,57 @@ describe('ExampleViewer', () => {
...
+    component.selectCorrectTab();
+    expect(component.selectedTab).toBe(1);
+
+    component.file = 'file.css';
+    component.selectCorrectTab();
+    expect(component.selectedTab).toBe(2);
+  }));
+
+  it('should generate correct url with region', async(() => {
+    fixture.detectChanges();
```
atscott, line 77: does it make sense to move this to the `beforeEach` rather than repeating in each test?

src/app/shared/example-viewer/example-viewer.spec.ts
```diff
@@ -48,12 +48,57 @@ describe('ExampleViewer', () => {
...
     component.toggleSourceView();
-    expect(component.view).toBe('collapsed');
+    expect(component.view).toBe('demo');
+  }));
+
+  it('should change to correct tab', async(() => {
+    fixture.detectChanges();
+    component.file = 'file.html';
+    component.selectCorrectTab();
+    expect(component.selectedTab).toBe(0);
```
atscott, line 65: Hmm, can we update this assert to be a bit more expressive. We actually don't actually care about the index number. What we care about is that when the file is `html`, we selected the `html` tab, regardless of what index that actually is. Can you use the tab harnesses to get the selected tab label? https://github.com/angular/components/tree/master/src/material/tabs/testing

PR #760, opened 2020-05-14
restructure doc items

annieyw:
Demo: https://mmalerba-demo2.firebaseapp.com/components

- documentation items: remove categories and sort alphabetically
- update side nav bar to fill page vertically
- replace categories page with 2d grid of all components


mmalerba:
small nit: it looks like the sidebar isn't quite tall enough:
image

jelbourn:
Could you constrain the text at the top to match the width of the card container?
image

src/app/pages/component-sidenav/_component-sidenav-theme.scss
```diff
@@ -43,6 +28,10 @@
         background: rgba(0, 0, 0, $nav-background-focus-opacity);
       }
     }
+
+    .mat-list-item {
+      color: rgba(mat-color($foreground, secondary-text), 0.6);
```
jelbourn, line 33: The contrast ratio here is a bit low. Was there a reason for changing the text color?

src/app/pages/component-sidenav/component-sidenav.scss
```diff
@@ -117,6 +86,14 @@ app-component-sidenav {
   flex: 1 1 auto;
 }
 
+.mat-list-base {
+  padding-top: 0 !important;
```
jelbourn, line 90: `!important` should be avoided; it's generally considered a major anti-pattern

src/app/pages/component-sidenav/_component-sidenav-theme.scss
annieyw: I thought the contrast was too high actually, was trying to go for something like this: https://standards.google/guidelines/google-material/components/
where the contrast is lower normally and increases on hover

jelbourn: This ratio is only 5.66, which doesn't meet the W3C's AAA level:
https://contrast-ratio.com/#rgba%280%2C0%2C0%2C.6%29-on-%23fafafa

The ratio on that internal site is 6.05

src/app/pages/component-category-list/component-category-list.html
```diff
@@ -2,12 +2,15 @@
      id="category-summary"
      focusOnNavigation>{{_categoryListSummary}}</div>
 <div class="docs-component-category-list">
-  <a *ngFor="let category of docItems.getCategories((params | async)?.section)"
+  <a *ngFor="let component of docItems.getItems((params | async)?.section)"
      class="docs-component-category-list-item"
-    [routerLink]="[category.id]">
+    [routerLink]="'/' + (params | async)?.section + '/' + component.id">
```
atscott, line 7:
```
<div class="docs-component-category-list" *ngIf="(params | async)?.section as section>
  <a *ngFor="let component of docItems.getItems(section)"
       class="docs-component-category-list-item"
      [routerLink]="'/' + section + '/' + component.id">
...
```
I think storing the (params | async)?.section could clean this up a bit

src/app/pages/component-sidenav/component-nav.html
```diff
@@ -1,23 +1,12 @@
...
-              [attr.aria-controls]="'panel-' + category.id"
-              [attr.aria-expanded]="panel.expanded">
-        {{category.name}}
-        <mat-icon>{{panel.expanded ? 'keyboard_arrow_up' :  'keyboard_arrow_down'}}</mat-icon>
-      </button>
-      <mat-nav-list dense [@bodyExpansion]="panel.expanded ? 'expanded' : 'collapsed'"
-                    id="panel-{{category.id}}" [attr.aria-label]="category.name">
-        <a mat-list-item *ngFor="let component of category.items"
+      <mat-nav-list>
+        <a mat-list-item *ngFor="let component of docItems.getItems((params | async)?.section)"
```
atscott, line 4: same comment here as above with creating a variable using ngIf

src/app/pages/component-sidenav/component-sidenav.ts
```diff
@@ -114,59 +113,18 @@ export class ComponentSidenav implements OnInit, OnDestroy {
...
 export class ComponentNav implements OnInit, OnDestroy {
   @Input() params: Observable<Params>;
-  expansions: {[key: string]: boolean} = {};
   currentItemId: string;
   private _onDestroy = new Subject<void>();
 
-  constructor(public docItems: DocumentationItems, private _router: Router) {}
+  constructor(public docItems: DocumentationItems) {}
 
   ngOnInit() {
```
atscott, line 121: remove if not needed

src/app/pages/component-sidenav/component-sidenav.ts
```diff
@@ -114,59 +113,18 @@ export class ComponentSidenav implements OnInit, OnDestroy {
 })
 export class ComponentNav implements OnInit, OnDestroy {
   @Input() params: Observable<Params>;
-  expansions: {[key: string]: boolean} = {};
   currentItemId: string;
   private _onDestroy = new Subject<void>();
```
atscott, line 117: looks like this isn't used.

src/app/pages/component-sidenav/_component-sidenav-theme.scss
annieyw: Changed to 0.65, now ratio is 6.86
https://contrast-ratio.com/#rgba%280%2C0%2C0%2C.65%29-on-%23fafafa

src/app/pages/component-sidenav/component-sidenav.scss
annieyw: I think the padding is part of the theme so that was the only way I could remove it. (same with `.mat-list-item-content` right below this)

src/app/pages/component-category-list/_component-category-list-theme.scss
```diff
@@ -8,4 +8,9 @@
   .docs-component-category-list-card-summary {
     color: mat-color($foreground, secondary-text);
   }
+
+  .mat-card {
+    box-shadow: none !important;
```
mmalerba, line 13: We shouldn't use `!important` (also card is pretty much just a div with `box-shadow` so now that we're turning off the `box-shadow` does it make sense to just use a div instead?

src/app/pages/component-sidenav/component-sidenav.scss
mmalerba: You don't need `!important` you just need a selector with higher specificity: https://developer.mozilla.org/en-US/docs/Web/CSS/Specificity

src/app/pages/component-viewer/component-viewer.scss
```diff
@@ -4,7 +4,7 @@ app-component-viewer {
   font-weight: 400;
 
   // spacing for the component viewer
-  padding: 20px 50px 50px 20px;
+  padding: 20px 50px 20px 50px;
```
mmalerba, line 7: nit: I think you can use the shorthand `padding: 20px 50px;`

src/app/pages/component-category-list/_component-category-list-theme.scss
annieyw: I changed it to a div. Also I feel that once we get the pictures in it'll make more sense for them to be in divs

src/app/pages/component-sidenav/component-sidenav.scss
```diff
@@ -117,6 +86,14 @@ app-component-sidenav {
   flex: 1 1 auto;
 }
 
+div .mat-list-base {
+  padding-top: 0;
+}
+
+.mat-list-item-content {
+  padding-left: 25px !important;
```
mmalerba, line 94: no `!important`

jelbourn:
Overall looks phenomenal. Here's some minor tweaks I would make, potentially in a follow-up PR.
* I would add a hover treatment for the items on the big components list and _maybe_ a ripple for clicks. 
* I would change the hover and selected treatment for the left nav:
  * For hover/focus, I would make the text full black instead of the primary color
  * For selected (active route), I would use a treatment with the primary color [like this one](https://lh3.googleusercontent.com/tVfI-EewJiKQqYvmcV8xkY2AIh_GXLSDlh5nzGka0SooqiJey_MWjtdpETNpBj8GEIzbE4LlaDIRX90ibVJlchUKKQ8qanhrRgFE=w1064-v0)
* For the text at the top:
  * We previously talked about whether we could make the text better align with the cards, and I said I'd think about it. I thought about it! I think we could use media queries to accomplish that [with something like this rough example](https://jsbin.com/cohojoj/1/edit?html,css,output)
  * I would increase the font size of that top text to 18px, `text-align: center`, and `line-height: 1.5`. I have no clue what the design principles behind this are, but this feels "cleaner" somehow. 
  * I would update the text (which I assume I originally wrote) to: 
    ```markdown    
    Angular Material offers a wide variety of UI components based on the [Material Design specification](https://material.io/components)
    ````
  * In a future change, I'd like to add a "Learn more" link at the end of the text and link to an "About" page where we go into more detail about both Angular Material and the CDK

PR #767, opened 2020-06-02
add button scene

annieyw:
First draft:
image


mmalerba:
I feel like these images might be too big overall, on my laptop I'd probably see 4 total components before having to scroll. 

I also think the disabled buttons don't look great in this example, I would suggest instead including some FABs (the circular buttons)

annieyw:
image

mmalerba:
@annieyw needs rebase

PR #768, opened 2020-06-02
add datepicker scene

annieyw:
First draft: 
datepicker scene


mmalerba:
I think this one is escaping your color filter because of the overlay. It might be easier to just use a `<mat-calendar>` instead of a full `<mat-datepicker>`.

annieyw:
image

mmalerba:
can we change the gray colors on the circles, days of the week, and dropdown arrow - they don't fit great with the color scheme

annieyw:
image

mmalerba:
What do the 2 white circles represent anyways? At first I thought it was the selected date, but then why are there 2 of them. That also reminds me, we should make one of these show a range selection, since we just added that feature

annieyw:
image

will update scene to use date range picker after 10.0

mmalerba:
@annieyw needs rebase

PR #769, opened 2020-06-02
add checkbox scene

annieyw:
First draft:
checkbox scene


mmalerba:
Can you get rid of the disabled checkbox? It doesn't look great. Also, is all of the text the same color here? I feel like every time I focus on one of them the others seem to look different, maybe it's an optical illusion...

annieyw:
image

I think the disabled checkbox text _might_ have been a different colour because now that it's gone they look the same.

scenes/src/app/scenes/checkbox/checkbox-scene.scss
```diff
@@ -0,0 +1,14 @@
+.scene-checkboxes {
+  width: 120px;
+  float: left;
```
mmalerba, line 3: can you make the host `display: flex` instead of using `float: left`

annieyw:
image

mmalerba:
@annieyw needs rebase

PR #770, opened 2020-06-03
add input scene

annieyw:
First draft:
input scene


mmalerba:
Can you use the fill and outline styles instead? Legacy and standard are no longer in the spec, I want to get rid of them eventually

annieyw:
image

atscott:
image
Hmmm, maybe just use a single column? It looks a bit weird with both being cut off.

mmalerba:
I'm trying to think how we'll distinguish the `input` image from the `form-field` image. The only thing I can really think of is to try to focus on variations of the input in this one vs variations of the form-field in the other one. e.g. maybe show some different `<input type="...">` (type=text, type=password, type=number) - a login screen with a pin would work for that.

The form field one could instead show the different appearances (fill and outline) as well as more hints/errors/prefix/suffix - to focus on things that are part of the form-field

annieyw:
image

mmalerba:
Can you remove the hint for the password? It makes things look pretty cluttered. 

Also, I wouldn't take the screenshot with the input focused - what we've found in screenshot tests is that its unreliable whether the caret is showing or not in the screenshot. We don't want it to randomly change every time we retake the screenshot.

annieyw:
image

scenes/e2e/src/app.e2e-spec.ts
```diff
@@ -1,11 +1,20 @@
...
   // they are not actually testing anything
   let page: AppPage;
+  const components = ['input'];
 
   beforeEach(() => {
     page = new AppPage();
   });
+
+  for (const comp of components) {
+    it('screenshot for ' + comp + ' scene', async () => {
```
mmalerba, line 15: nit: same comment as other PR about using backtick string (can just wait for that one to merge then rebase, accept their edit, and put 'input' back in the list above)

scenes/src/app/scenes/input/input-scene.scss
```diff
@@ -0,0 +1,21 @@
+.scene-inputs .scene-input-clear {
+  margin-bottom: -10px
```
mmalerba, line 2: nit end with ;

PR #771, opened 2020-06-03
add progress-spinner scene

annieyw:
First draft:
image


scenes/src/app/scenes/progress-spinner/progress-spinner-scene.scss
```diff
@@ -0,0 +1,12 @@
+mat-progress-spinner {
+  position: absolute;
```
atscott, line 2: Is it possible to use flexbox or some other alignment strategy to get this perfectly centered rather than trying to calculate it?

scenes/src/styles.scss
```diff
@@ -30,4 +30,5 @@ $theme: map_merge($theme, (foreground: $foreground));
 body {
   margin: 0;
   color: red;
+  font-family: Roboto, Arial, sans-serif;
```
atscott, line 33: Should this change get submitted separately if it's going to be needed for many scenes?

mmalerba:
can you try throwing in an indeterminate spinner too? image feels sparse

annieyw:
image

now that the image is smaller adding another spinner made the image really crowded so I opted for a "shadow" that shows remaining progress

PR #776, opened 2020-06-03
add ripples scene

annieyw:
First draft:
image


mmalerba:
I like the looks of this one, but now that I think about it more, you should be able to do it with actual ripples. They have a persistent option you can use to make them stay open. Here's an example: https://stackblitz.com/edit/angular-n6fgfl?file=src/app/ripple-overview-example.ts

annieyw:
recreated scene with actual ripples
image

scenes/src/app/scenes/ripples/ripples-scene.html
```diff
@@ -0,0 +1,9 @@
+<div class="scene-ripples-wrapper" #wrapper
+  matRipple
+  [matRippleDisabled]="true">
+</div>
+<button mat-raised-button color="primary" id="scene-ripples-button" #button
+        matRipple
+        [matRippleDisabled]="true">RIPPLES</button>
+
+
```
mmalerba, line 9: nit: extra lines

scenes/e2e/src/app.e2e-spec.ts
```diff
@@ -1,11 +1,20 @@
...
   // they are not actually testing anything
   let page: AppPage;
+  const components = ['ripples'];
 
   beforeEach(() => {
     page = new AppPage();
   });
+
+  for (const comp of components) {
+    it('screenshot for ' + comp + ' scene', async () => {
```
mmalerba, line 15: nit: I always prefer backtick strings when doing this kind of thing:
```ts
`screenshot for ${comp} scene`
```

PR #778, opened 2020-06-04
fix: right side overflow

wszgrcy:
like this [table](https://material.angular.io/components/table/examples)
the right side overflow becase of the example


wszgrcy:
@googlebot I signed it!

Splaktar:
Thank you for your contribution and for reporting this issue. Unfortunately, this needs to be fixed in the example, which can be found in https://github.com/angular/components/blob/59d39840440d1dd35767de5290fec10960f583f5/src/components-examples/material/table/table-sticky-columns/table-sticky-columns-example.css.

I've opened PR https://github.com/angular/components/pull/20063 to fix this.

PR #780, opened 2020-06-05
landing page redesign

annieyw:
- 1st row: general info
- 2nd row: horizontal guides carousel 
- 3rd row: (weekly?) top components

demo: https://material2-docs-dev.web.app/

### DIFF
--- a/src/app/pages/homepage/_homepage-theme.scss
+++ b/src/app/pages/homepage/_homepage-theme.scss
@@ -26,7 +26,7 @@
       color: mat-color($foreground, secondary-text);
     }
 
-    .docs-homepage-featured-components h2 {
+    .docs-homepage-carousel-row h2 {
       color: mat-color($foreground, secondary-text);
     }
 
@@ -46,11 +46,19 @@
       .docs-homepage-guides-card {
         background: mat-color($background, hover);
       }
+
+      img {
+        opacity: .5;
+      }
     }
 
     [carousel-item]:focus {
       .docs-homepage-guides-card {
         background: mat-color($background, focused-button);
+      }
+
+      img {
+        transform: scale(1.2);
       }
       outline: none;
     }
--- a/src/app/pages/homepage/_homepage-theme.scss
+++ b/src/app/pages/homepage/_homepage-theme.scss
@@ -16,10 +16,21 @@
 
     .docs-header-start {
       color: mat-color($primary);
+
+      .docs-link {
+        color: mat-color($primary, default-contrast);
+      }
     }
 
     .docs-homepage-row {
       color: mat-color($foreground, secondary-text);
     }
+
+    .docs-homepage-bottom-start {
+      .docs-link {
+        color: mat-color($foreground, text)
+      }
+    }
+
   }
 }
--- a/src/app/pages/homepage/_homepage-theme.scss
+++ b/src/app/pages/homepage/_homepage-theme.scss
@@ -26,11 +26,11 @@
       color: mat-color($foreground, secondary-text);
     }
 
-    .docs-homepage-featured-components h1 {
+    .docs-homepage-featured-components h2 {
       color: mat-color($foreground, secondary-text);
     }
 
-    a {
+    a.docs-link, a.docs-featured-components-carousel-item {
       color: mat-color($foreground, secondary-text)
     }
 
@@ -43,13 +43,13 @@
     }
 
     [carousel-item]:hover {
-      .docs-guides-card.mat-card {
+      .docs-homepage-guides-card {
         background: mat-color($background, hover);
       }
     }
 
     [carousel-item]:focus {
-      .docs-guides-card.mat-card {
+      .docs-homepage-guides-card {
         background: mat-color($background, focused-button);
       }
       outline: none;
--- a/src/app/pages/homepage/_homepage-theme.scss
+++ b/src/app/pages/homepage/_homepage-theme.scss
@@ -46,21 +46,29 @@
       .docs-homepage-guides-card {
         background: mat-color($background, hover);
       }
-
-      img {
-        opacity: .5;
-      }
     }
 
     [carousel-item]:focus {
       .docs-homepage-guides-card {
         background: mat-color($background, focused-button);
       }
+      outline: none;
+    }
 
+    .docs-featured-components-carousel-item:hover {
       img {
         transform: scale(1.2);
       }
-      outline: none;
+
+      box-shadow:inset 0 0 0 1.5px mat-color($foreground, divider);
+    }
+
+    .docs-featured-components-carousel-item:focus {
+      img {
+        transform: scale(1.2);
+      }
+
+      box-shadow:inset 0 0 0 1.5px mat-color($foreground, disabled);
     }
   }
 }
--- a/src/app/pages/homepage/_homepage-theme.scss
+++ b/src/app/pages/homepage/_homepage-theme.scss
@@ -26,7 +26,7 @@
       color: mat-color($foreground, secondary-text);
     }
 
-    .docs-link {
+    a {
       color: mat-color($foreground, secondary-text)
     }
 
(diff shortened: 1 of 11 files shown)


atscott:
Some quick/initial usage notes:
* The carousel horizontal scroll is a little hard to use, especially on mac. Do we have a real carousel or can we find one to use?
* We might want to think about the order of the guides in that list. Schematics are nice, but I don't think they're more important than theming (so should come later). I also think the component harnesses could be given some more love in priority.
* Maybe swap the guides row with the top components row?
* I kind of wanted to click on the top components to get more info. Could we have them link to the component?
* It doesn't scale super well for smaller screens. If I make my browser window smaller or use the device emulator of devtools, the view is a little messed up.
image

* <3 those component scenes! I really like having some images of the components on the home page.

src/app/pages/homepage/homepage.html
```diff
@@ -6,55 +6,81 @@ <h1 class="mat-h1">Angular Material</h1>
...
-      <docs-svg-viewer src="../assets/img/homepage/sprintzerotoapp.svg"
-                       [scaleToContainer]="true"></docs-svg-viewer>
+  <div class="docs-homepage-row docs-introduction">
+    <div class="docs-homepage-row-column">
+      <h2>High quality</h2>
+      <p>
+        Internationalized and accessible components for everyone. Well tested to ensure performance and reliability.
+      </p>
+      <p>Straightforward APIs with consistent cross platform behaviour.</p>
     </div>
```
annieyw, line 22: @jelbourn please feel free to add/suggest actually meaningful text here, I got most of this from the README of the components repo

annieyw:
- top row with 3 columns turn into 3 rows when screen is small
- featured components in carousel
updated demo: https://material2-docs-dev.web.app/

src/app/pages/homepage/_homepage-theme.scss
```diff
@@ -16,10 +16,43 @@
...
 
     .docs-homepage-row {
       color: mat-color($foreground, secondary-text);
     }
+
+    .docs-homepage-featured-components h1 {
+      color: mat-color($foreground, secondary-text);
+    }
+
+    a {
```
mmalerba, line 33: Do we really want this to be all `<a>` elements or should it be `a.docs-link`?

src/app/pages/homepage/_homepage-theme.scss
```diff
@@ -16,10 +16,43 @@
...
+    a.docs-link:hover {
+      color: mat-color($foreground, text);
+    }
+
+    .docs-homepage-guides-card-divider {
+      background: mat-color($accent)
+    }
+
+    [carousel-item]:hover {
+      .docs-guides-card.mat-card {
```
mmalerba, line 46: Is the extra specificity here necessary? (i.e. could we just do `.docs-guides-card {`)

src/app/pages/homepage/homepage.scss
```diff
@@ -1,26 +1,43 @@
...
+$margin-promotion-sections: 50px;
 $margin-promotion-sections-small: 15px;
 
 .docs-header-background {
   overflow: hidden;
+  position: relative;
+}
+
+.docs-header-background::before {
+  content: " ";
```
mmalerba, line 11: I think it can just be: `content: '';` (empty string, no space). Also use single quote

src/app/pages/homepage/homepage.scss
```diff
@@ -1,26 +1,43 @@
...
+  position: absolute;
+  background-image: url("../../../assets/img/homepage/angular-white-transparent.svg");
```
mmalerba, line 13: Use single quote

src/app/pages/homepage/homepage.html
```diff
@@ -6,55 +6,84 @@ <h1 class="mat-h1">Angular Material</h1>
...
-    <div class="docs-homepage-promo-desc">
-      <h2>Versatile</h2>
-      <p>Themeable, for when you need to stay on brand or just have a favorite color.
-        Accessible and internationalized so that all users are welcome.</p>
-    </div>
+
+  <mat-divider></mat-divider>
+
+  <div class="docs-homepage-featured-components">
+    <h1>Featured components</h1>
```
atscott, line 46: nit: can this be an h2? I think the jury is still out on whether pages should only have a single h1 tag

src/app/pages/homepage/homepage.html
```diff
@@ -6,55 +6,84 @@ <h1 class="mat-h1">Angular Material</h1>
       <h2> Material Design components for Angular</h2>
```
atscott, line 6: While we're here, should we update the page header/sub-header?

annieyw:
Add guides overview in guides carousel, add hover/focus effect for featured components

src/app/pages/homepage/homepage.html
```diff
@@ -6,60 +6,81 @@ <h1 class="mat-h1">Angular Material</h1>
...
-      Get started
+
+  <mat-divider></mat-divider>
+
+  <div class="docs-homepage-featured-components docs-homepage-carousel-row">
+    <h2>Featured components</h2>
+    <app-carousel [itemWidth]="260" [aria-label]="'Featured components'">
+      <a *ngFor="let comp of getTopComponents()" carousel-item class='docs-featured-components-carousel-item'
+         routerLink="/components/{{comp}}">
+        <div class="docs-homepage-img-container"><img alt="{{comp}}" src="../../../assets/screenshots/{{comp}}.scene.png"></div>
```
mmalerba, line 50: Can you put `role="presentation"` on the image? Currently ChromeVox is reading everything twice, once for the image and once for the text. e.g. "button button"

src/app/shared/carousel/carousel.html
```diff
@@ -1,13 +1,12 @@
-<div class="docs-carousel-content-wrapper" #contentWrapper (window:resize)="onResize()"
-     [style.minWidth]="shiftWidth + 'px'">
-  <button mat-mini-fab class="docs-carousel-nav docs-carousel-nav-prev" (click)="previous()" *ngIf="this.showPrevArrow">
+<div #contentWrapper (keyup)="onKeydown($event)"
+     (window:resize)="onResize()" [style.minWidth]="shiftWidth + 'px'" class="docs-carousel-content-wrapper" role="region">
+  <button (click)="previous()" *ngIf="this.showPrevArrow" class="docs-carousel-nav docs-carousel-nav-prev" mat-mini-fab>
```
mmalerba, line 3: the previous and next buttons both need an `aria-label`

src/app/shared/carousel/carousel.ts
```diff
@@ -10,17 +10,23 @@ import {
...
+export class CarouselItem implements FocusableOption {
+  @HostBinding('attr.role') readonly role = 'listitem';
   @HostBinding('style.width.px') width = this.carousel.itemWidth;
   @HostBinding('tabindex') tabindex = '-1';
 
   constructor(readonly carousel: Carousel, readonly element: ElementRef) {
   }
+
+  focus(): void {
+    this.element.nativeElement.focus({preventScroll: true});
```
mmalerba, line 28: Why prevent scroll? I just tested and found that I could sometimes get the focus to be off screen

annieyw: The automatic scrolling conflicts with `_shiftItems` when you get to the end of the visible cards and the next focus is not yet visible, do you know if theres a way to just disable horizontal scroll?

src/app/pages/homepage/homepage.html
annieyw: I tried putting `role="presentation"` but ChromeVox was still reading everything twice so I just removed the `alt` . I think this should be fine since it's same as the label below

src/app/shared/carousel/carousel.ts
jelbourn: I would make the images `aria-hidden="true"` since they're purely decorative

Issue #257, opened 2017-09-02
bug(docs): toc is not sticky


donroyco:
Bug.

**Mozilla Firefox: _Not sticky_**
When navigation in Mozilla Firefox version 55.0.2 on a [components page](https://material.angular.io/components/icon/overview) while scrolling down, the table of contents won't stay at the top of the viewport, but just scrolls along with the rest of the content.

![toc-scroll](https://user-images.githubusercontent.com/1763537/29995337-338c04c2-8fe7-11e7-93a8-8c83f7e03a9a.gif)

**Apple Safari: _Bottom of page_**
In Apple Safari, version 10.1.1, the table of contents is at the bottom of the page (below screenshot) and also not sticky, but this may be a side effect of #247 | #248 

![toc-safari](https://user-images.githubusercontent.com/1763537/29995356-95e696dc-8fe7-11e7-8190-b27fa5ac5329.png)

**Browser**
Mozilla Firefox, version 55.0.2
Apple Safari, version 10.1.1

**What else?**
No errors in console.

amcdnl:
In Safari, you can toggle various different parent elements display properties and it starts working. For instance, toggling this ![img](https://content.screencast.com/users/amcdaniel22/folders/Snagit/media/42a4935a-767e-49cd-b044-8ecf15f3c13e/2017-09-02_09-43-12.png) in the dev tools makes it work, however, if you set it via css or even programmatically in JS, it doesn't work. I'm guessing this is a reflow issue.

donroyco:
@amcdnl Thanks for looking into this.

annieyw:
Fixed in https://github.com/oktav777/material.angular.io/pull/1

Splaktar:
It looks like PR https://github.com/angular/material.angular.io/pull/362 is still open to solve this. Though it hasn't been rebased in a few years. I'm not sure if it's still relevant/needed?

Splaktar:
I closed #362 since similar changes were made in PR https://github.com/angular/material.angular.io/pull/840, but they did not fix the sticky TOC issues.

Splaktar:
@tommyc38 is correct in https://github.com/angular/material.angular.io/issues/916#issue-801031281 that `overflow: hidden` in the component-sidenav.scss is what is breaking the sticky TOC.

[...]

PR #784, opened 2020-06-08
follow-up improvements to components list page

annieyw:
follow up of https://github.com/angular/material.angular.io/pull/760

- change color/hover of side nav
- hover effect on cards
- text wrap constraint on cards

### DIFF
--- a/src/app/pages/component-sidenav/component-sidenav.ts
+++ b/src/app/pages/component-sidenav/component-sidenav.ts
@@ -49,7 +49,7 @@ import {NavigationFocusService} from '../../shared/navigation-focus/navigation-f
 // These breakpoint values need to stay in sync with the related Sass variables in
 // src/styles/_constants.scss.
 const EXTRA_SMALL_WIDTH_BREAKPOINT = 720;
-const SMALL_WIDTH_BREAKPOINT = 959;
+const SMALL_WIDTH_BREAKPOINT = 939;
 
 @Component({
   selector: 'app-component-sidenav',
--- a/src/app/shared/documentation-items/documentation-items.ts
+++ b/src/app/shared/documentation-items/documentation-items.ts
@@ -43,8 +43,8 @@ const COMPONENTS = 'components';
 export const SECTIONS: { [key: string]: DocSection } = {
   [COMPONENTS]: {
     name: 'Components',
-    summary: 'Angular Material comprises a range of components which implement common ' +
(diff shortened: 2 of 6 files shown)


annieyw:
Before:
image

After:
image

I think lightening the text makes it a little better

src/app/shared/documentation-items/documentation-items.ts
```diff
@@ -48,11 +48,8 @@ export const SECTIONS: { [key: string]: DocSection } = {
   },
   [CDK]: {
     name: 'CDK',
-    summary: 'The Component Dev Kit (CDK) is a set of tools that implement common interaction ' +
-      'patterns whilst being unopinionated about their presentation. It represents an ' +
-      'abstraction of the core functionalities found in the Angular Material library, without ' +
-      'any styling specific to Material Design. Think of the CDK as a blank state of well-tested ' +
-      'functionality upon which you can develop your own bespoke components.'
+    summary: 'The Component Dev Kit (CDK) is a set of behavior primitives for building UI' +
```
mmalerba, line 51: I mean this fixes it for sufficiently large screens... but it's still gonna look weird on smaller ones. I think the right way to do this if you want it centered is to make a wrapper flexbox that centers its item and allow that item to expand as necessary to fill the container, but the text in the item is still aligned left like normal

src/app/pages/component-sidenav/component-sidenav.ts
```diff
@@ -49,7 +49,7 @@ import {NavigationFocusService} from '../../shared/navigation-focus/navigation-f
 // These breakpoint values need to stay in sync with the related Sass variables in
 // src/styles/_constants.scss.
 const EXTRA_SMALL_WIDTH_BREAKPOINT = 720;
-const SMALL_WIDTH_BREAKPOINT = 959;
+const SMALL_WIDTH_BREAKPOINT = 939;
```
Splaktar, line 52: I think that it's problematic that this small breakpoint was changed so that it no longer matches our Sass small breakpoint:

https://github.com/angular/material.angular.io/blob/53965ba3b8d271e4b7b29a8923f2c2ddba5dc488/src/styles/_constants.scss#L11

It also no longer matches any of the established Material Design layout breakpoints:
https://material.io/design/layout/responsive-layout-grid.html#breakpoints

I'm looking at reverting this in https://github.com/angular/material.angular.io/issues/257.

PR #785, opened 2020-06-09
add scale input to scene viewer

annieyw:
scale scenes instead of overwriting specific dimensions


mmalerba:
looks like there's a failing lint check, otherwise good to go

PR #786, opened 2020-06-10
horizontal carousel for guide list on landing page

annieyw:


src/app/shared/horizontal-carousel/horizontal-carousel.spec.ts
```diff
@@ -0,0 +1,49 @@
...
+import {DocsAppTestingModule} from '../../testing/testing-module';
+
+describe('HorizontalCarousel', () => {
+  let fixture: ComponentFixture<HorizontalCarousel>;
+  let component: HorizontalCarousel;
+
+  beforeEach(async(() => {
+    TestBed.configureTestingModule(
+      {imports: [HorizontalCarouselModule, DocsAppTestingModule]}
+    ).compileComponents();
```
atscott, line 12: These tests pass if I remove `compileComponents` (and the `async`). I think because the components are aot compiled, all the async resources are inlined already so this isn't needed.

// cc @mmalerba

src/app/shared/horizontal-carousel/horizontal-carousel.spec.ts
```diff
@@ -0,0 +1,49 @@
...
+    expect(component.index).toEqual(1);
+
+    const navPrevious = fixture.nativeElement.querySelector('.docs-carousel-nav-prev');
+    expect(navPrevious).toBeDefined();
+  });
+
+  it('should trigger onResize method when window is resized', () => {
+    const spyOnResize = spyOn(component, 'onResize');
+    window.dispatchEvent(new Event('resize'));
+    expect(spyOnResize).toHaveBeenCalled();
```
atscott, line 37: Hmm, on its own, I don't think this test has much value. I think it would be good to combine it with the one below and add the desired data to the event. Additionally, I think you can then make `resizeCarousel` private

src/app/shared/horizontal-carousel/horizontal-carousel.ts
```diff
@@ -0,0 +1,88 @@
...
+
+@Component({
+  selector: 'app-horizontal-carousel',
+  templateUrl: './horizontal-carousel.html',
+  styleUrls: ['./horizontal-carousel.scss'],
+  encapsulation: ViewEncapsulation.None,
+})
+
+export class HorizontalCarousel implements AfterViewInit {
+  position: number = 0;
```
atscott, line 19:
```suggestion
  position = 0;
```

src/app/shared/horizontal-carousel/horizontal-carousel.ts
```diff
@@ -0,0 +1,88 @@
...
+  showPrevArrow = false;
+  showNextArrow = true;
+  guideItemsLen: number;
+  visibleCards: number;
+
+  constructor(public guideItems: GuideItems) {
+    this.guideItemsLen = guideItems.getAllItems().length;
```
atscott, line 26: nit: I would kind of prefer that `Length` is spelled out... But I won't fight this.

src/app/shared/horizontal-carousel/horizontal-carousel.ts
```diff
@@ -0,0 +1,88 @@
...
+})
+
+export class HorizontalCarousel implements AfterViewInit {
+  position: number = 0;
+  showPrevArrow = false;
+  showNextArrow = true;
+  guideItemsLen: number;
+  visibleCards: number;
+
+  constructor(public guideItems: GuideItems) {
```
atscott, line 25:
```suggestion
  constructor(readonly guideItems: GuideItems) {
```

src/app/shared/horizontal-carousel/horizontal-carousel.ts
atscott, line 26:
```suggestion
  guideItemsLen = this.guideItems.getAllItems().length;
  visibleCards: number;

  constructor(public guideItems: GuideItems) {
```

src/app/shared/horizontal-carousel/horizontal-carousel.ts
```diff
@@ -0,0 +1,88 @@
...
+      carousel.style.width = `${this.visibleCards * CARD_WIDTH}px`;
+    }
+  }
+
+  next() {
+    this.index += 1;
+    const cards = document.getElementsByClassName('docs-guides-card');
+    this.position += CARD_WIDTH;
+    [].forEach.call(cards, (card: HTMLElement) => {
+      card.style.transform = `translateX(-${this.position}px)`;
```
atscott, line 67: I think the translate actually needs to go on the anchor elements. If you tab through them, you'll get some weird behavior because the anchor isn't moved: 
image

src/app/shared/horizontal-carousel/horizontal-carousel.ts
```diff
@@ -0,0 +1,88 @@
...
+import {MatIconModule} from '@angular/material/icon';
+import {MatButtonModule} from '@angular/material/button';
+
+const CARD_WIDTH = 250;
+
+@Component({
+  selector: 'app-horizontal-carousel',
+  templateUrl: './horizontal-carousel.html',
+  styleUrls: ['./horizontal-carousel.scss'],
+  encapsulation: ViewEncapsulation.None,
```
atscott, line 15: It looks like maybe you're using `viewEncapsulation.None` so the buttons are aligned correctly? It would be great if we could remove this.

What about having this component be `display: flex; flex-direction: row` and then having the buttons be hidden instead of removing them with `*ngIf`? That way they have a reserved space and the carousel won't jump when they show/hide.

src/app/shared/horizontal-carousel/horizontal-carousel.ts
```diff
@@ -0,0 +1,88 @@
...
+    [].forEach.call(cards, (card: HTMLElement) => {
+      card.style.transform = `translateX(-${this.position}px)`;
+    });
+  }
+
+  prev() {
+    this.index -= 1;
+    const cards = document.getElementsByClassName('docs-guides-card');
+    this.position -= CARD_WIDTH;
+    [].forEach.call(cards, (card: HTMLElement) => {
```
atscott, line 75: hmm, this is weird. Why not just `cards.forEach`? (same question above)

src/app/shared/horizontal-carousel/horizontal-carousel.ts
```diff
@@ -0,0 +1,88 @@
...
+  next() {
+    this.index += 1;
+    const cards = document.getElementsByClassName('docs-guides-card');
+    this.position += CARD_WIDTH;
+    [].forEach.call(cards, (card: HTMLElement) => {
+      card.style.transform = `translateX(-${this.position}px)`;
+    });
+  }
+
+  prev() {
```
atscott, line 71:
```suggestion
  previous() {
```

atscott:
Oh, one more thing I thought of - the left/right buttons aren't very visible. Is there something we can do to make them more discoverable?
There are some carousels on SERP that you might be able to check out:
image
This one is Youtube Music in dark mode:
image

src/app/shared/horizontal-carousel/horizontal-carousel.ts
mmalerba: I'm assuming this is because getElementsByClassName doesn't return an array, you can do this:
```ts
const cards = [...document.querySelectorAll('.docs-guides-card')];
for (const card of cards) {...}
```

src/app/shared/horizontal-carousel/horizontal-carousel.scss
```diff
@@ -0,0 +1,50 @@
...
+  margin: 10px;
+  padding: 20px;
+  text-align: center;
+}
+
+.docs-carousel-card-container {
+  transition: transform 0.5s ease-in-out;
+}
+
+.mat-card-title {
```
annieyw, line 32: @mmalerba none of the `mat-card` styles are applying

jelbourn:
I haven't look at anything, but I'll just leave this here: https://inclusive-components.design/a-content-slider/

src/app/shared/horizontal-carousel/horizontal-carousel.ts
```diff
@@ -0,0 +1,103 @@
...
+    this._index = i;
+    this.showPrevArrow = i > 0;
+    this.showNextArrow = i < (this.items.length - this.visibleCards);
+  }
+
+  onResize() {
+    this._resizeCarousel();
+  }
+
+  ngAfterContentInit(): void {
```
atscott, line 60: I know it doesn't happen now in the current use-case, but what about using [cdkObserveContent](https://material.angular.io/cdk/observers/api#ContentObserver) and calling `resizeCarousel` if the contents change? Then you could support adding/removing items in the carousel. I guess this might be feature creep now :)

src/app/shared/horizontal-carousel/horizontal-carousel.ts
```diff
@@ -0,0 +1,103 @@
...
+
+  previous() {
+    this.index -= 1;
+    this.position -= this.shiftWidth;
+    this.items.forEach((card: CarouselItem) => {
+      card.elem.nativeElement.style.transform = `translateX(-${this.position}px)`;
+    });
+  }
+
+  private _resizeCarousel() {
```
atscott, line 83: I don't think this is a big deal: if I make the screen really small, scroll to the end, then expand the screen again, I only see one item all the way on the left. It _could_ be nice to scroll some others back into view if there's space available. Not necessary though; the scrolling still works as expected.

src/app/shared/horizontal-carousel/horizontal-carousel.ts
```diff
@@ -0,0 +1,103 @@
...
+  NgModule,
+  Optional,
+  QueryList,
+} from '@angular/core';
+import {CommonModule} from '@angular/common';
+import {MatIconModule} from '@angular/material/icon';
+import {MatButtonModule} from '@angular/material/button';
+import {CarouselItem} from './horizontal-carousel-directive';
+
+const CAROUSEL_WIDTH_RATIO = 0.75; // carousel width should be approximately 75% of page
```
atscott, line 18: Hmm, this seems a _little_ specific. The carousel only works when it's a specific width? Are we able to use the native element and get the actual width instead with offsetWidth/clientWidth/innerWidth?

src/app/shared/horizontal-carousel/horizontal-carousel-directive.ts
```diff
@@ -0,0 +1,13 @@
...
+import {HorizontalCarousel} from './horizontal-carousel';
+
+@Directive({
+  selector: '[carousel-item]',
+})
+export class CarouselItem {
+  @HostBinding('style.width') width = `${this.carousel.itemWidth}px`;
+  @HostBinding('style.height') height = `${this.carousel.itemHeight}px`;
+
+  constructor(public carousel: HorizontalCarousel, public elem: ElementRef) {
```
atscott, line 11:
```suggestion
  constructor(readonly carousel: HorizontalCarousel, readonly elem: ElementRef) {
```
nit: use readonly

src/app/pages/homepage/homepage.ts
```diff
@@ -1,20 +1,25 @@
...
   templateUrl: './homepage.html',
-  styleUrls: ['./homepage.scss']
+  styleUrls: ['./homepage.scss'],
+  encapsulation: ViewEncapsulation.None
 })
 export class Homepage implements OnInit {
   isNextVersion = location.hostname.startsWith('next.material.angular.io');
 
-  constructor(public _componentPageTitle: ComponentPageTitle) {}
+  constructor(public _componentPageTitle: ComponentPageTitle, public guideItems: GuideItems) {}
```
atscott, line 22:
```suggestion
  constructor(readonly _componentPageTitle: ComponentPageTitle, readonly guideItems: GuideItems) {}
```

src/app/shared/horizontal-carousel/horizontal-carousel.ts
```diff
@@ -0,0 +1,103 @@
...
+  }
+
+  ngAfterContentInit(): void {
+    setTimeout(() => {
+      this.shiftWidth = this.items.first.elem.nativeElement.clientWidth;
+      this._resizeCarousel();
+    });
+  }
+
+  next() {
```
atscott, line 67: One use-ability note: I can tab to items that aren't visible on the carousel. This results in some odd behavior (carousel scrolls to the right but the component doesn't know so the scrolling breaks). Can we do something about that? Maybe set the tab index to -1?

annieyw:
Switched to adjusting size based on width of host instead of window. Looking into cdkObserveContent

src/app/pages/homepage/homepage.html
```diff
@@ -55,6 +55,23 @@ <h2>Optimized for Angular</h2>
       <p>Built by the Angular team to integrate seamlessly with Angular.</p>
     </div>
   </div>
+
+  <div style="width: 75%">
```
mmalerba, line 59: add a class and apply this in the CSS file instead

src/app/pages/homepage/homepage.html
```diff
@@ -55,6 +55,23 @@ <h2>Optimized for Angular</h2>
...
+  <app-carousel [itemWidth]="250" [itemHeight]="250">
```
mmalerba, line 60: I don't think the carousel needs to enforce a height on the content, it can just expand to fit them

src/app/pages/homepage/homepage.html
```diff
@@ -55,6 +55,23 @@ <h2>Optimized for Angular</h2>
...
+    <a carousel-item class='docs-carousel-item-container' *ngFor="let guide of guideItems.getAllItems()"
```
mmalerba, line 61: can we call the class `docs-carousel-item`? `docs-carousel-item-container` makes it sound like a wrapper around this element

src/app/pages/homepage/homepage.scss
```diff
@@ -86,6 +86,31 @@ $margin-promotion-sections-small: 15px;
...
+
+.docs-guides-card.mat-card {
+  text-align: center;
+}
+
+mat-card-title {
+  font-size: 20px;
+}
+
+a {
```
mmalerba, line 110: be more specific, just using `a` may cause this style to get applied to elements you didn't intend for it to affect

src/app/pages/homepage/homepage.scss
```diff
@@ -86,6 +86,31 @@ $margin-promotion-sections-small: 15px;
...
+  transition: transform 0.5s ease-in-out;
+  padding: 10px;
+  display: flex;
+}
+
+.docs-guides-card.mat-card {
+  text-align: center;
+}
+
+mat-card-title {
```
mmalerba, line 106: this should be more specific or it will affect all `mat-card-title`

src/app/shared/carousel/carousel-directive.ts
```diff
@@ -0,0 +1,14 @@
+import {Directive, ElementRef, HostBinding} from '@angular/core';
+import {Carousel} from './carousel';
+
+@Directive({
+  selector: '[carousel-item]',
+})
+export class CarouselItemDirective {
+  @HostBinding('style.width') width = `${this.carousel.itemWidth}px`;
```
mmalerba, line 8: Instead I think you can do `@HostBinding('style.width.px') width = this.carousel.itemWidth`

src/app/shared/carousel/carousel.html
```diff
@@ -0,0 +1,14 @@
+<button mat-mini-fab class="docs-carousel-nav docs-carousel-nav-prev" (click)="previous()"
+        [style.visibility]="this.showPrevArrow ? 'visible' : 'hidden'">
+  <mat-icon>navigate_before</mat-icon>
+</button>
+<div class="docs-carousel-wrapper" (window:resize)="onResize()" #wrapper [style.minWidth]="shiftWidth + 'px'">
```
mmalerba, line 5: can we call it `docs-carousel-content-wrapper`?

src/app/shared/carousel/carousel.html
```diff
@@ -0,0 +1,14 @@
...
+  <div class="docs-carousel">
```
mmalerba, line 6: `docs-carousel-content`?

src/app/shared/carousel/carousel.html
```diff
@@ -0,0 +1,14 @@
+<button mat-mini-fab class="docs-carousel-nav docs-carousel-nav-prev" (click)="previous()"
+        [style.visibility]="this.showPrevArrow ? 'visible' : 'hidden'">
```
mmalerba, line 2: can you use `*ngIf` to hide instead of `style.visibility`?

src/app/shared/carousel/carousel.scss
```diff
@@ -0,0 +1,16 @@
+:host {
+  display: flex;
```
mmalerba, line 2: do all of these actually need to be `display:flex`?

src/app/shared/carousel/carousel.html
mmalerba: and `#contentWrapper`

src/app/shared/carousel/carousel-directive.ts
```diff
@@ -0,0 +1,14 @@
+import {Directive, ElementRef, HostBinding} from '@angular/core';
+import {Carousel} from './carousel';
+
+@Directive({
+  selector: '[carousel-item]',
+})
+export class CarouselItemDirective {
```
mmalerba, line 7: can call the class just `CarouselItem` and the file `carousel-item.ts`

src/app/shared/carousel/carousel.ts
```diff
@@ -0,0 +1,100 @@
...
+    this.index += shiftIndex;
+    this.position += shiftIndex * this.shiftWidth;
+    this.items.forEach((item: CarouselItemDirective) => {
+      item.elem.nativeElement.style.transform = `translateX(-${this.position}px)`;
+    });
+  }
+
+  private _resizeCarousel() {
+    const newVisibleCards = Math.max(1, Math.min(
+      Math.floor((this.host.nativeElement.offsetWidth - 2 * BUTTON_WIDTH) / this.shiftWidth),
```
mmalerba, line 79: should you measure the content wrapper instead of the host so you don't have to hardcode button sizes?

src/app/shared/carousel/carousel.ts
```diff
@@ -0,0 +1,100 @@
...
+      }
+      this.visibleCards = newVisibleCards;
+      this.showNextArrow = this.index < (this.items.length - this.visibleCards);
+    }
+    this.visibleCards = newVisibleCards;
+    this.wrapper.nativeElement.style.width = `${this.visibleCards * this.shiftWidth}px`;
+  }
+}
+
+@NgModule({
```
mmalerba, line 94: you can move this to a separate file `carousel-module.ts`

annieyw:
hide buttons using `*ngIf`
slide content back when resized
demo: https://material2-docs-dev.web.app/

src/app/shared/carousel/carousel.html
```diff
@@ -0,0 +1,14 @@
+<div class="docs-carousel-content-wrapper" #contentWrapper (window:resize)="onResize()"
+     [style.minWidth]="shiftWidth + 'px'">
+  <button mat-mini-fab class="docs-carousel-nav docs-carousel-nav-prev" (click)="previous()" *ngIf="this.showPrevArrow">
+    <mat-icon>navigate_before</mat-icon>
+  </button>
+  <div class="docs-carousel-content">
```
atscott, line 6: Maybe add left + right arrow keyboard support as done [here](https://inclusive-components.design/a-content-slider/#keyboardsupport).
```suggestion
  <div class="docs-carousel-content" tabindex="0" (keydown.arrowright)="next()" (keydown.arrowleft)="previous()">
```

src/app/shared/carousel/carousel.spec.ts
```diff
@@ -0,0 +1,103 @@
...
+  });
+});
+
+@Component({
+  selector: 'test-carousel',
+  template:
+      `
+    <app-carousel itemWidth="250">
+      <div carousel-item class="docs-carousel-item-container"
+           *ngFor="let i of [].constructor(numberOfItems) "></div>
```
atscott, line 95: Hmmm, this might require a sizeable refactor so maybe add a follow up to investigate:
The content slider accessibility guide's [checklist](https://inclusive-components.design/a-content-slider/#checklist) suggested using " list markup to group the slides together. Then screen reader users in 'browse' mode can use list navigation shortcuts to traverse them."

I think you can inspect the styles on the ul/li elements in this [demo](https://codepen.io/heydon/full/xPWOLp) or what's [listed in the guide](https://inclusive-components.design/a-content-slider/#slides) to get it working.

```
ul {
  display: flex;
}

li {
  list-style: none;
  flex: 0 0 100%;
}
```

annieyw:
- arrow key nav support when focused on carousel
- calculate visible items and set `tabindex` accordingly 
updated demo (https://material2-docs-dev.web.app/)

src/app/shared/carousel/carousel.spec.ts
annieyw: I tried implementing this but it got a bit messy so I think I will do this in a followup PR along with the cdkObserveContent

src/app/shared/carousel/carousel.ts
```diff
@@ -0,0 +1,121 @@
...
+
+  previous() {
+    // prevent keyboard navigation from going out of bounds
+    if (this.showPrevArrow) {
+      this._shiftItems(-1);
+      this.setTabIndex();
+    }
+  }
+
+  private setTabIndex() {
```
mmalerba, line 88: maybe call it `updateItemTabIndecies`, the current name makes it sound like a method to set a tabindex on the host. also can it be private?

annieyw:
addressed nits and removed demo from homepage

PR #794, opened 2020-06-22
add radio button scene

annieyw:
image


scenes/src/app/scenes/radio/radio-scene.html
```diff
@@ -0,0 +1,8 @@
+<div class="scene-radio-title">
+  <mat-icon class="scene-radio-icon">language</mat-icon>
+</div>
```
atscott, line 3: I think this scene would look fine without this, but I don't feel strongly either way.

scenes/src/app/scenes/radio/radio-scene.scss
```diff
@@ -0,0 +1,18 @@
...
+  margin: 5px;
+}
+
+.scene-radio-title {
+  text-align: center;
+}
+
+.scene-radio-icon {
+  font-size: 45px;
+  color: rgba(red, 0.6);
```
atscott, line 17: Is this hardcoded color going to be an issue if the hue changes?

scenes/src/app/scenes/radio/radio-scene.html
mmalerba: +1 I would prefer not to have the icon

scenes/src/app/scenes/radio/radio-scene.scss
mmalerba: All of the examples are actually red, but then we use a CSS filter to shift the hue, so I think this should be ok

annieyw:
image

Issue #333, opened 2017-11-21
When printing a docs page, the print layout is chopped off after 1 page


mmalerba:
Originally filed as https://github.com/angular/material2/issues/8378

The issue here is that the body is not the element that scrolls in desktop mode. Its a similar problem to making the address bar scroll out of the way on mobile, and we can probably reuse some of the styles we use for mobile mode for print mode

philip-firstorder:
I fixed this and the font error below by adding some print css in _general.scss

<img width="894" alt="roboto print problem" src="https://user-images.githubusercontent.com/30979663/43611209-a9b0da7e-96a8-11e8-8983-2cc8c3d1ef08.png">

Here is the code that fixes all print problems on Chrome, Safari, Firefox, Edge and Opera:
Demo: https://firstorder.io/imprint
GitHub: https://github.com/firstorder-gmbh/firstorder.io/blob/master/src/styles/_general.scss

```scss
@media print {
  // Force to print colors, otherwise the material components will look white
  color-adjust: exact;
  -webkit-print-color-adjust: exact;

  body {
    // Roboto not printing ok on Chrome, fallback on supported fonts
    font-family: 'Helvetica Neue', sans-serif !important;
    
    mat-sidenav-container {
      // Remove background color to save toner
      background-color: transparent !important;
      // Override the color from the template
      color: black !important;
      // Position needs to be relative
      position: relative !important;

      mat-sidenav-content {
        // Remove the content margin if sidenav is expanded when hidden for print
        margin: 0 !important;

        // Remove max width and padding from containers to flow text depending on printer page sizes
        .container {
          min-width: 0 !important;
          padding: 0 !important;
        }

        // Roboto not printing ok on Chrome, fallback on supported fonts
        .mat-typography, .mat-card {
          font-family: 'Helvetica Neue', sans-serif !important;
          h1, h2, h3, h4, h5 {
            font-family: 'Helvetica Neue', sans-serif !important;
          }
        }
        .mat-button, .mat-raised-button, .mat-icon-button,
        .mat-stroked-button, .mat-flat-button, .mat-fab, .mat-mini-fab {
          font-family: 'Helvetica Neue', sans-serif !important;
        }

        // Define where a page break should occur when printing
        h2 {
          page-break-before: always; // always break before h2 elements, for chapters for example
        }
        h3, h4 {
          page-break-after: avoid; // avoid breaking after subtitles, it looks weird
        }
        pre, blockquote, img {
[...]

PR #795, opened 2020-06-23
add badge scene

annieyw:
image


scenes/src/app/scenes/badge/badge-scene.ts
```diff
@@ -0,0 +1,21 @@
...
+  selector: 'app-badge-scene',
+  templateUrl: './badge-scene.html',
+  styleUrls: ['./badge-scene.scss']
+})
+export class BadgeScene {
+}
+
+@NgModule({
+  imports: [
+    DemoMaterialModule
```
mmalerba, line 15: Can't we just import the badge and icon modules? the demo one has a _lot_ of stuff

annieyw: I think it auto imported the wrong one oops

scenes/src/app/scenes/badge/badge-scene.scss
```diff
@@ -0,0 +1,4 @@
+.scene-badge-icon {
+  color: palevioletred;
```
mmalerba, line 2: oh I think this is why I was trippin about the colors, what if you use a dark gray instead? maybe something close to `#333`?

annieyw:
image

scenes/src/app/scenes/badge/badge-scene.scss
```diff
@@ -0,0 +1,4 @@
+.scene-badge-icon {
+  color: #333333;
```
mmalerba, line 2: nit: You can use the 3 character shortcut `#333` (this shortcut works when each color channel is just the same digit twice - e.g. `#abc === #aabbcc`)

annieyw:
Changed to `#333`

mmalerba:
@annieyw needs rebase

PR #798, opened 2020-06-24
center skip link

annieyw:
skip link was off centered 
`left: 44%` looked fine on most laptop screens but won't be centered on bigger/smaller screens

### DIFF
--- a/src/app/shared/navbar/navbar.scss
+++ b/src/app/shared/navbar/navbar.scss
@@ -43,7 +43,8 @@
 .skip-link-wrapper {
   position: absolute;
   top: 10px;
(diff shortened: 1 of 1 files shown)


src/app/shared/navbar/navbar.scss
```diff
@@ -41,9 +41,11 @@
 }
 
 .skip-link-wrapper {
+  width: 180px;
```
mmalerba, line 44: what is this `width` for?

annieyw: I think accidentally left it in there when I was trying something

PR #10, opened 2016-11-11
chore: add ci setup

DevVersion:
**In Progress**

PR #799, opened 2020-06-24
Update to Angular 10.0.0

jelbourn:


package.json
```diff
@@ -25,33 +25,33 @@
...
-    "@angular/cdk": "10.0.0-next.1",
-    "@angular/cdk-experimental": "10.0.0-next.1",
-    "@angular/common": "^10.0.0-rc.0",
-    "@angular/compiler": "^10.0.0-rc.0",
+    "@angular/animations": "10.0.0",
+    "@angular/cdk": "10.0.0",
+    "@angular/cdk-experimental": "10.0.0",
+    "@angular/common": "^10.0.0",
+    "@angular/compiler": "^10.0.0",
     "@angular/components-examples": "angular/material2-docs-content#master",
```
devversion, line 33: Should probably use `#10.0.x`?

devversion: I assume there is not a lot of change in the docs-content for stuff that only landed in `angular/components#master`, but might be worth double-checking, or just using the `10.0.x` branch.

jelbourn:
Updated stackblitz and docs-content sha, PTAL

PR #802, opened 2020-06-25
set scale on container

annieyw:
- scale style wasn't setting properly after adding new `scene-content-container` div.
- scale for divider scene was overridden at some point  
- typo in `ripple`


mmalerba:
do the screenshots need to be updated or there was no change?

annieyw:
Some of the screenshots were never checked in/didn't have the final screenshot checked in so I added them.

PR #803, opened 2020-06-26
fix(guides): fix the positioning of the footer on the guides page

andy9775:
On the cdk and components pages the footer always displays below the bottom of the viewport whereas
on the guides page it displays in the middle of the page when viewed on larger devices. This fix
ensures that the guides page takes up 100% of the given viewport and the footer is displayed
outside the viewport matching the other pages.

----- 
example of current functionality 
image


mmalerba:
Nice, even wrote a test for it! LGTM

PR #806, opened 2020-06-30
update code snippet view

annieyw:
Code snippets on the doc site use Roboto Mono with `font-weight: 300` this caused it to be hard to read on Windows.
image

- add `font-weight: 400` for Roboto Mono
- change border around static snippets to visually separate them from surrounding text and to be consistent with live snippets

`font-weight: 400` without border change:
image

`font-weight: 400` with border change:
image

dark mode:
Screen Shot 2020-06-30 at 3 10 15 PM

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -49,7 +49,7 @@
   <meta name="theme-color" content="#3f51b5">
 
   <link href="https://fonts.googleapis.com/icon?family=Material+Icons&display=block" rel="stylesheet">
-  <link href="https://fonts.googleapis.com/css?family=Roboto:300,400,500,700|Roboto+Mono:300,400&display=swap" rel="stylesheet">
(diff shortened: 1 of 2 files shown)


src/index.html
```diff
@@ -49,7 +49,7 @@
   <meta name="theme-color" content="#3f51b5">
 
   <link href="https://fonts.googleapis.com/icon?family=Material+Icons&display=block" rel="stylesheet">
-  <link href="https://fonts.googleapis.com/css?family=Roboto:300,400,500,700|Roboto+Mono:300&display=swap" rel="stylesheet">
+  <link href="https://fonts.googleapis.com/css?family=Roboto:300,400,500,700|Roboto+Mono:300,400&display=swap" rel="stylesheet">
```
mmalerba, line 52: do we need both? is the 300 level still used anywhere?

Splaktar: I doubt that there's any reason to use Mono 300 anywhere.

annieyw:
removed 300, updated with dark mode screenshot

PR #807, opened 2020-06-30
add dialog scene

annieyw:
image


atscott:
Hmm, this one's a bit hard to tell that it's a dialog? Can we make it a little smaller than the width of the scene so you can see all the borders?

mmalerba:
also instead of making the text green, can we make the dialog background green?

annieyw:
image

PR #814, opened 2020-07-01
List scene

nielsr98:


scenes/e2e/src/app.e2e-spec.ts
```diff
@@ -26,6 +26,7 @@ describe('screenshot scenes for each component', () => {
     'slider',
     'stepper',
     'snack-bar',
+    'list',
```
annieyw, line 29: nit: keep list in alphabetical order to avoid merge conflicts

PR #823, opened 2020-07-06
add form-field scene

annieyw:
image


scenes/src/app/scenes/form-field/form-field-scene.html
```diff
@@ -0,0 +1,8 @@
+<mat-form-field appearance="outline">
+  <mat-label>Password</mat-label>
+  <input matInput [type]="hide ? 'password' : 'text'" required>
+  <mat-hint align="end">Forgot password?</mat-hint>
+  <button mat-icon-button matSuffix (click)="hide = !hide" [attr.aria-label]="'Hide password'" [attr.aria-pressed]="hide">
```
atscott, line 5: Can this be removed? It doesn't look like anything changes the value.

scenes/src/app/scenes/form-field/form-field-scene.html
```diff
@@ -0,0 +1,6 @@
+<mat-form-field appearance="outline">
```
Splaktar, line 1: We previously decided that the `fill` appearance would be what we used as the default in the docs. We should probably use that here and have the screen shot reflect that.

annieyw:
image

PR #826, opened 2020-07-07
add scene screenshots to components list

annieyw:
demo: https://material2-docs-dev.web.app/components/categories
- tweaks to some scenes
- update screenshots
- optimize png with `pngquant`
- add scene screenshots to components list


src/app/pages/component-category-list/component-category-list.scss
```diff
@@ -8,27 +8,33 @@
...
-    max-width: 220px;
+    max-width: 600px;
+  }
+}
+
+@media (max-width: 694px) {
+  .docs-component-category-list-summary {
+    max-width: 280px;
   }
 }
```
atscott, line 40: I wonder if it would look better if we centered the cards with `justify-content: center` in `.docs-component-category-list`.

I think you would also need to center the summary text, which looks like it would require `display: flex;
    flex-direction: column;` on the `app-component-category-list`

image

annieyw:
updated demo: https://material2-docs-dev.web.app/components/categories
- update button scene text from "BUTTON" to "Button"
- clean up imports etc in scenes
- center cards and text

PR #840, opened 2020-07-21
fix(material-docs-app): remove absolute positioning of all content

Splaktar:
- preserve behavior of not having a scrollbar next to the toolbar
- hide theme picker on small devices (iPhone 5) so that navbar doesn't wrap

Screen Shot 2020-07-21 at 18 51 33


src/app/material-docs-app.scss
```diff
@@ -20,21 +16,14 @@ material-docs-app > app-component-sidenav {
   flex: 1;
 }
 
-material-docs-app > app-homepage,
-material-docs-app > app-guides,
-material-docs-app > guide-viewer {
+material-docs-app > router-outlet + * {
```
mmalerba, line 19: selectors that end with `*` aren't the most efficient, is there any other way to do this?

src/app/shared/navbar/navbar.scss
```diff
@@ -69,3 +69,10 @@
     margin: 0;
   }
 }
+theme-picker {
```
mmalerba, line 72: nit: newline

src/styles/_general.scss
```diff
@@ -1,3 +1,7 @@
+html,
+body {
+  height: 100vh;
+}
```
mmalerba, line 4: nit: newline

src/app/material-docs-app.scss
Splaktar: I don't think so as these are all going to be custom components, some that start with `app-` (like `app-guides`) and some that don't (like `guide-viewer`. Hopefully the `+` limits the scope of the matching for `*` enough that this isn't too inefficient.

Splaktar:
Conflicts resolved.

src/app/material-docs-app.scss
mmalerba: Selectors are parsed right-to-left, so the `+` won't help. Is there some way we could apply a common class to all of them instead?

Splaktar:
The missing 2 points from the LH a11y audit is fixed in PR https://github.com/angular/components/pull/21750.

Splaktar:
The other 2 missing points from the LH a11y audit for the Getting Started Guide are fixed in PR https://github.com/angular/components/pull/21751.

src/app/material-docs-app.scss
Splaktar: Done

Splaktar:
OK, this should be ready to go now.

mmalerba (approved):
LGTM

PR #845, opened 2020-07-25
docs(button): Exclude toggle button examples from button examples.

walvekarnikhil:
* Examples are matched using the prefix specified in documentation-items.ts
* Button & button-toggle both start with 'button-' prefix, so examples are getting mixed.
* We will exclude certain examples using "exclude" property.

### DIFF
--- a/src/app/shared/documentation-items/documentation-items.ts
+++ b/src/app/shared/documentation-items/documentation-items.ts
@@ -89,7 +89,7 @@ const DOCS: { [key: string]: DocItem[] } = {
       summary: 'An interactive button with a range of presentation options.',
       exampleSpecs: {
         prefix: 'button-',
-        exclude: ['button-toggle-appearance', 'button-toggle-exclusive', 'button-toggle-overview']
+        exclude: ['button-toggle-']
       },
       additionalApiDocs: [{name: 'Testing', path: 'material-button-testing.html'}],
     },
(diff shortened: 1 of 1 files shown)


walvekarnikhil:
Fixes: [#19533](https://github.com/angular/components/issues/19533)

Toggle button examples are not shown now.
Screenshot 2020-07-25 at 2 49 56 PM

src/app/shared/documentation-items/documentation-items.ts
```diff
@@ -89,6 +89,7 @@ const DOCS: { [key: string]: DocItem[] } = {
       summary: 'An interactive button with a range of presentation options.',
       exampleSpecs: {
         prefix: 'button-',
+        exclude: ['button-toggle-appearance', 'button-toggle-exclusive', 'button-toggle-overview']
```
mmalerba, line 92: @annieyw would it work if he just did `exclude: ['button-toggle-']`?

walvekarnikhil: That may not work since we are checking if the example name is present in the exclude list (array includes).
```
  doc.examples =
    exampleNames
      .filter(key => key.match(RegExp(`^${doc.exampleSpecs.prefix}`)) &&
        !doc.exampleSpecs.exclude?.includes(key));
```

For `exclude: ['button-toggle-']` to work we could change the code to:
```
  doc.examples =
    exampleNames
      .filter(key => key.match(RegExp(`^${doc.exampleSpecs.prefix}`)) &&
        !doc.exampleSpecs.exclude?.every(excludeName => key.startsWith(excludeName)));
```

mmalerba: Do you want to make that change then? Seems more robust.

I think it should be `some` instead of `every` though. And IIRC not all browsers support `startsWith` so `indexOf(excludeName) === 0` might be a better way to go

annieyw: Changing it to 
```
doc.examples =
    exampleNames
      .filter(key => key.match(RegExp(`^${doc.exampleSpecs.prefix}`)) &&
        !doc.exampleSpecs.exclude?.every(excludeName => key.startsWith(excludeName)));
```
should work

PR #848, opened 2020-08-03
 Use `img` element in component list w/ lazy loading

jelbourn:
Switch the component list to use the `img` element so we can take advantage of native lazy-loading.


mmalerba:
Have you run a demo to make sure this doesn't cause any jumpiness as the images load in? I think this CSS property can be removed now https://github.com/jelbourn/material.angular.io/blob/lazy-images/src/app/pages/component-category-list/component-category-list.scss#L61

jelbourn:
Removed that css property. Here's the deployed dev app with the change: https://material2-docs-dev.firebaseapp.com/components/categories. I haven't seen any jumpiness; the images do visibly load after scrolling if you go super fast, but otherwise it's pretty seamless with the native `lazy` behavior.

PR #856, opened 2020-08-14
fix: router outlet should not allow overflow

andrewseguin:
Fixes https://github.com/angular/components/issues/20313


Splaktar:
I actually turned down this same fix in https://github.com/angular/material.angular.io/pull/778.

PR #861, opened 2020-09-07
build: update to Angular 10.0

crisbeto:
* Updates to Angular 10.1 and TypeScript 4.0.
* Replaces the custom `CopierService` with the `Clipboard` service from the CDK.


devversion (approved):
LGTM. Nice cleanup of the clipboard copier 🎉

PR #866, opened 2020-09-16
build: fix angular cli commonjs warnings for polyfills

devversion:
The polyfills file was incorrectly imported through the `main` file
of the CLI project. This caused undesired CommonJS dependency
warnings for ZoneJS. We fix this by removing the import of the
polyfills and letting the CLI handle the polyfills as normally done
in fresh CLI projects.

Also disables the zone-error import from the devmode environment.
ZoneJS generally has the potential of hiding errors (as seen with
the `Object.defineProperty` patch that was enabled by default in
non-evergreen Zone bundles), so we should disable it. It seems
better aligning devmode as much as possible with prodmode to catch
issues more quickly. Debugging without `zone-error` is not any
different from common Angular/ZoneJS errors as seen in the components
or framework repositories.

### DIFF
--- a/angular.json
+++ b/angular.json
@@ -20,6 +20,7 @@
             "sourceMap": true,
             "outputPath": "dist/material-angular-io",
             "index": "src/index.html",
+            "polyfills": "src/polyfills.ts",
             "main": "src/main.ts",
             "tsConfig": "tsconfig.app.json",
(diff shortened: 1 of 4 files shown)


Splaktar (approved):
LGTM after rebase.

PR #867, opened 2020-09-21
update to 10.2.2

wagnermaciel:


Splaktar:
It looks like this can be closed since https://github.com/angular/material.angular.io/pull/886 updated to `11.0.0`.

PR #878, opened 2020-10-20
fix(toc): visibility on resizing

geromegrignon:
## Motivation

On the documentation page of a component, resizing down the width under 1200px makes the right sided table of contents disappears. By resizing the width to more than 1000px, the table of contents does not re-appear.

## Reproduction

Navigate to any component documentation from the [overview](https://material.angular.io/components/categories).
Resize the width to less than 1200px.
Resize it back to more than 1200px.

=> the table of contents isn't present.

## Changes

add change detection
trigger the event emitters to populate the toc

Closes [angular/components issue ](https://github.com/angular/components/issues/20315)

### DIFF
--- a/src/app/pages/component-viewer/component-viewer.ts
+++ b/src/app/pages/component-viewer/component-viewer.ts
@@ -1,24 +1,29 @@
 import {BreakpointObserver} from '@angular/cdk/layout';
 import {CommonModule} from '@angular/common';
 import {
+  ChangeDetectorRef,
   Component,
   Directive,
   NgModule,
   OnDestroy,
   OnInit,
+  QueryList,
   ViewChild,
+  ViewChildren,
   ViewEncapsulation,
 } from '@angular/core';
 import {MatTabsModule} from '@angular/material/tabs';
 import {ActivatedRoute, Params, Router, RouterModule} from '@angular/router';
 import {combineLatest, Observable, ReplaySubject, Subject} from 'rxjs';
-import {map, takeUntil} from 'rxjs/operators';
+import {map, skip, takeUntil} from 'rxjs/operators';
 import {DocViewerModule} from '../../shared/doc-viewer/doc-viewer-module';
 import {DocItem, DocumentationItems} from '../../shared/documentation-items/documentation-items';
 import {TableOfContents} from '../../shared/table-of-contents/table-of-contents';
 import {TableOfContentsModule} from '../../shared/table-of-contents/table-of-contents.module';
 import {ComponentPageTitle} from '../page-title/page-title';
 import {NavigationFocusModule} from '../../shared/navigation-focus/navigation-focus';
+import {DocViewer} from '../../shared/doc-viewer/doc-viewer';
+
 
 @Component({
   selector: 'app-component-viewer',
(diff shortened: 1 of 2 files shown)


geromegrignon:
For some reason, the **Files changed** show an empty line added into the constructors. It isn't present in the changed code, as visible into my [fork](https://github.com/geromegrignon/material.angular.io/blob/toc-hidden/src/app/pages/component-viewer/component-viewer.ts).

src/app/pages/component-viewer/component-viewer.ts
```diff
@@ -90,6 +102,18 @@ export class ComponentBaseView implements OnInit, OnDestroy {
...
+
+    this.showToc.pipe(
+      skip(1),
+      takeUntil(this.destroyed)
+    ).subscribe(() => {
+      if (this.tableOfContents) {
+        this.viewers.forEach(viewer => {
+          viewer.contentRendered.emit(viewer._elementRef.nativeElement);
+        });
+        this.tableOfContents.updateScrollPosition();
```
Splaktar, line 114: I don't think that this line is required since `contentRendered` triggers `updateTableOfContents()` or `updateScrollPosition()` in all cases. This just causes it to be called twice.

src/app/pages/component-viewer/component-viewer.ts
```diff
@@ -74,14 +79,21 @@ export class ComponentViewer implements OnDestroy {
...
-  constructor(public componentViewer: ComponentViewer, breakpointObserver: BreakpointObserver) {
+  constructor(
+    public componentViewer: ComponentViewer,
+    breakpointObserver: BreakpointObserver,
+    private changeDetectorRef: ChangeDetectorRef) {
     this.showToc = breakpointObserver.observe('(max-width: 1200px)')
-      .pipe(map(result => !result.matches));
+      .pipe(
+        map(result => !result.matches),
+        tap(() => this.changeDetectorRef.detectChanges())
```
Splaktar, line 95:
```suggestion
        map(result => {
          this.changeDetectorRef.detectChanges();
          return !result.matches;
        })
```

src/app/pages/component-viewer/component-viewer.ts
```diff
@@ -1,24 +1,29 @@
...
+  QueryList,
   ViewChild,
+  ViewChildren,
   ViewEncapsulation,
 } from '@angular/core';
 import {MatTabsModule} from '@angular/material/tabs';
 import {ActivatedRoute, Params, Router, RouterModule} from '@angular/router';
 import {combineLatest, Observable, ReplaySubject, Subject} from 'rxjs';
-import {map, takeUntil} from 'rxjs/operators';
+import {map, skip, takeUntil, tap} from 'rxjs/operators';
```
Splaktar, line 18:
```suggestion
import {map, skip, takeUntil} from 'rxjs/operators';
```

Splaktar (changes requested):
Thank you for your contribution!

We can trigger the CD in the `map()` rather than adding the `tap` operator and calling it in that.

I don't like the extra `detectChanges()` but they do appear to be required to fix this. I did verify the fix works, with and without my suggestions.

src/app/pages/component-viewer/component-viewer.ts
geromegrignon: I removed that line.

src/app/pages/component-viewer/component-viewer.ts
geromegrignon: I added your suggestion.

src/app/pages/component-viewer/component-viewer.ts
geromegrignon: I removed the import.

geromegrignon:
Thanks for your review @Splaktar.
I committed the suggested changes.

Splaktar (approved):
LGTM

PR #891, opened 2020-11-21
Docs titles clipped on the left side - anchors not accessible

abhishek2303:
Fixes the visibility of the anchor titles documentation UI which is currently clipped. This is done by assigning CSS `overflow: visible` property to the parent component - component-overview.

Fixes [20687](https://github.com/angular/components/issues/20687)

_**Before**_
Screenshot 2020-11-22 at 4 54 14 AM

_**After**_
Screenshot 2020-11-22 at 4 54 47 AM

The approach mentioned in [20687](https://github.com/angular/components/issues/20687) suggests removing the margin of -30px from `header-link`. This would make the header text start after 30px with nothing in between. The approach in this PR is taken to overcome this visual oddity.

### DIFF
--- a/src/app/pages/component-viewer/component-viewer.scss
+++ b/src/app/pages/component-viewer/component-viewer.scss
@@ -29,6 +29,7 @@ app-component-viewer {
   component-overview, component-api {
     display: flex;
     align-items: flex-start;
+    overflow: visible;
 
     @media (max-width: $small-breakpoint-width) {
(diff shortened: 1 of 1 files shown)


abhishek2303:
@googlebot I signed it!

crisbeto (approved):
LGTM. I also confirmed that this doesn't regress https://github.com/angular/material.angular.io/pull/856.

PR #900, opened 2020-12-30
fix: use new Google Fonts snippet

crisbeto:
Updates to the newer version of the Google Fonts snippet.

Related to https://github.com/angular/components/pull/21479.

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -48,11 +48,10 @@
   <meta name="msapplication-config" content="assets/img/favicons/browserconfig.xml?v=8.2.3">
   <meta name="theme-color" content="#3f51b5">
 
+  <link rel="preconnect" href="https://fonts.gstatic.com">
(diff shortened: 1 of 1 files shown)


src/index.html
```diff
@@ -48,11 +48,10 @@
   <meta name="msapplication-config" content="assets/img/favicons/browserconfig.xml?v=8.2.3">
   <meta name="theme-color" content="#3f51b5">
 
+  <link rel="preconnect" href="https://fonts.gstatic.com">
+  <link href="https://fonts.googleapis.com/css2?family=Roboto+Mono&family=Roboto:wght@300;400;500;700&display=swap" rel="stylesheet">
```
Splaktar, line 52: We previously only loaded weight 400 of Roboto Mono. Does `family=Roboto+Mono` load only 400 weight by default? Does the size of this stay the same or decrease?

crisbeto: I wasn't really looking at the size, but the stylesheet that it serves up looked the same. I didn't write the URL manually, it's what Google Fonts generated when I selected Roboto Mono 400 through the UI.

jelbourn (approved):
LGTM

PR #901, opened 2021-01-04
fix: generated Stackblitz examples not working when forked to a repo

crisbeto:
We weren't generating a full valid HTML file when creating the `index.html` for Stackblitz which meant that if somebody creates a repository out of it, it won't work. These changes resolve the issue by providing the full HTML file.

Providing the full HTML also allows us to represent a real app better and to fix some issues like the one where the app background doesn't cover the whole page.

Fixes https://github.com/angular/components/issues/21461.

### DIFF
--- a/src/assets/stack-blitz/src/index.html
+++ b/src/assets/stack-blitz/src/index.html
@@ -1,5 +1,13 @@
-<link href="https://fonts.googleapis.com/icon?family=Material+Icons&display=block" rel="stylesheet">
-<div class="mat-app-background basic-container">
-  <material-docs-example>loading</material-docs-example>
-</div>
-<span class="version-info">Current build: {{version}}</span>
+<!DOCTYPE html>
+<html lang="en">
+<head>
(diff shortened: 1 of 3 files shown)


src/assets/stack-blitz/src/index.html
```diff
@@ -1,5 +1,13 @@
-<link href="https://fonts.googleapis.com/icon?family=Material+Icons&display=block" rel="stylesheet">
-<div class="mat-app-background basic-container">
-  <material-docs-example>loading</material-docs-example>
-</div>
-<span class="version-info">Current build: {{version}}</span>
+<!DOCTYPE html>
+<html lang="en">
+<head>
+  <meta charset="UTF-8">
```
Splaktar, line 4:
```suggestion
  <meta charset="utf-8">
```
For consistency with what's generated via `ng new`.  This value is case-insensitive .

Splaktar (approved):
LGTM

PR #902, opened 2021-01-04
feat: add 404 page

crisbeto:
Sets up a dedicated 404 page, rather than redirecting everything to the front page which doesn't tell the user that their link is incorrect. The design is based on the one from angular.io.

I also had to make some adjustments to the routing setup, because everything was being picked up by the `:section` route and we didn't have defined routes for `/cdk` and `/components` to which we link from the top nav.

Angular_Material_UI_component_library_-_Google_Chr_2021-01-04_18-07-41

### DIFF
--- a/src/app/pages/not-found/not-found.html
+++ b/src/app/pages/not-found/not-found.html
@@ -1,0 +1,30 @@
+<main>
+  <div class="wrapper">
+    <svg
+      version="1.1"
+      xmlns="http://www.w3.org/2000/svg"
+      xmlns:xlink="http://www.w3.org/1999/xlink"
+      viewBox="0 0 200 200"
+      xml:space="preserve">
+    <path class="shield-left" d="M5.7,33.2L98.8,0l95.5,32.6l-15.4,123.1L98.8,200L20,156.3L5.7,33.2z"/>
+    <path class="shield-right" d="M194.3,32.6L98.8,0v200l80.1-44.3L194.3,32.6L194.3,32.6z"/>
+    <circle class="eye" cx="61.7" cy="80" r="10.7"/>
+    <circle class="eye" cx="138.3" cy="80" r="10.7"/>
+    <path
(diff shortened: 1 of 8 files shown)


src/app/pages/not-found/not-found.scss
```diff
@@ -0,0 +1,32 @@
...
+main {
+  min-height: 100vh;
+  font-size: 1.25rem;
+}
+
+.wrapper {
+  display: flex;
+  align-items: center;
+  justify-content: center;
+  padding: $spacing;
```
Splaktar, line 12: I prefer a bit more padding on top (64px)
Screen Shot 2021-01-11 at 21 00 00

Versus the 32px used here:
Screen Shot 2021-01-11 at 20 59 34

src/app/pages/not-found/not-found.html
```diff
@@ -0,0 +1,28 @@
...
+      class="frown"
+      stroke-width="10"
+      stroke-linecap="round"
+      fill="none"
+      d="M138,130.6c0,0-33.5-42.5-76,0"/>
+    </svg>
+
+    <div>
+      <h1>Page Not Found</h1>
+      <p>We're sorry. The page you are looking for cannot be found.</p>
```
Splaktar, line 23: I think that it would be helpful to include a couple anchor buttons here to help take them somewhere useful.
- 'Go Home' - `'/'`
- 'Read Guides' - `'/guides'`

AIO has both relevant search results (often empty) and some common links. We don't have search, but linking to some common top level sections of the app would be helpful.

crisbeto: I've added the "Go home" and "Read guides" for now and we can iterate on showing more content once it's in master.

crisbeto:
The feedback has been addressed.

Splaktar (approved):
LGTM

PR #907, opened 2021-01-22
build: update to Angular 11.1.0

crisbeto:
Bumps all of the Angular dependencies to 11.1.0 and TS 4.1.2.


package.json
```diff
@@ -83,6 +83,6 @@
     "shelljs": "^0.8.4",
     "ts-node": "^8.10.2",
     "tslint": "^6.1.3",
-    "typescript": "4.0.2"
+    "typescript": "4.1.2"
```
Splaktar, line 86: `ng update` doesn't bump this, but I saw that it's supported. I guess we'll try it out and see if there are any issues.

package.json
```diff
@@ -83,6 +83,6 @@
     "shelljs": "^0.8.4",
```
Splaktar, line 83: `ng update` also bumps up to `"karma": "~5.2.3",`

Splaktar (approved):
LGTM with one comment about karma.

package.json
crisbeto: I bumped it since it's the version we're using on the main repo.

PR #912, opened 2021-01-29
fix(component-viewer): Loading document... getting wrapped

Splaktar:
- when there is plenty of room, the message is being wrapped due to the width
  of the new parent span that was added in #717

# Before

Screen Shot 2021-01-29 at 14 36 15

# After

Screen Shot 2021-01-29 at 14 35 53

Harness API docs aren't side-by-side
Screen Shot 2021-01-29 at 14 58 40

### DIFF
--- a/src/app/pages/component-viewer/component-api.html
+++ b/src/app/pages/component-viewer/component-api.html
@@ -4,22 +4,22 @@
   </span>
 
   <!--
-  The span container is necessary for the layout to display properly. When
+  This container is necessary for the layout to display properly. When
   a component has multiple API docs, they need to be joined together in the
   same container so that they display one after another.
   -->
-  <span>
+  <div class="docs-component-api">
     <doc-viewer [documentUrl]="getApiDocumentUrl(docItem!)"
-      class="docs-component-view-text-content docs-component-api"
+      class="docs-component-view-text-content"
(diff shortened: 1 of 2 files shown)


src/app/pages/component-viewer/component-viewer.scss
```diff
@@ -59,7 +59,6 @@ app-component-viewer {
 .docs-component-api,
 .docs-component-overview {
   width: 80%;
-  display: inline-flex;
```
Splaktar, line 62: This fixes the side-by-side API docs when there are also harness API docs.

src/app/pages/component-viewer/component-api.html
```diff
@@ -4,22 +4,22 @@
   </span>
 
   <!--
-  The span container is necessary for the layout to display properly. When
+  This container is necessary for the layout to display properly. When
   a component has multiple API docs, they need to be joined together in the
   same container so that they display one after another.
   -->
-  <span>
+  <div class="docs-component-api">
```
Splaktar, line 11: This causes the container to start with 80% width instead of something like 50px width that wraps the message.

I changed it from a `span` since the layout needs to be block and not inline.

annieyw (approved):

PR #931, opened 2021-03-12
release: update next.material.angular.io with v12

annieyw:
future v11 doc site updates should be done on branch `11.x`


Splaktar (commented):
Sorry for the delay in looking at this. We'll want to bump the StackBlitz stuff (in the writer and `assets/stack-blitz`) to v12 and add v12 to the `assets/versions.json` as well.

Splaktar:
I can post a PR to update those in a few minutes.

PR #932, opened 2021-03-16
build: run ng update migrations for v12-next

Splaktar:
- update StackBlitz for v12-next
- add v12-next to version list


annieyw:
I think the stackblitz update is messing with the examples:
https://angularmaterial.dev/components/autocomplete/examples

Splaktar:
Doh, that doesn't look good. Thank you for the heads up. I'll take a look later today after some meetings.

Splaktar:
I looked into this some more and the autocomplete examples are broken on https://next.material.angular.io/components/autocomplete/examples as well. So this isn't an issue introduced with this PR, but I will look into resolving it.

Splaktar:
So this is looking like either Angular core is broken or some edge case of how we dynamically load components is broken. I'm going to open a separate issue for this in `angular/components` since it seems related to how we generate `component-examples`.

Splaktar:
Here's the issue to track these form-field example issues: https://github.com/angular/components/issues/22285

Splaktar:
OK, next.5 CLI build deployed as well. This should be ready to merge.

andrewseguin (approved):

PR #947, opened 2021-04-14
build: update to node 14 and remove dependency

crisbeto:
Updates the CI to Node 14 to match the rest of the organization.
Also removes the dependency on `angular-in-memory-web-api` from the generated Stackblitz since it isn't being used.


devversion (approved):

Splaktar (commented):
LGTM!

PR #949, opened 2021-04-16
fix(csp): allow img-src of 'self' for v11.x

Splaktar:
- needed for `data:image/png;base64` images used in some examples

Fixes:
Screen Shot 2021-04-15 at 22 33 27


wagnermaciel (approved):
lgtm

PR #952, opened 2021-04-16
feat: switch to sass modules

crisbeto:
Switches all of the .scss files to the Sass module system and to use the new Material theming API.

### DIFF
--- a/src/app/pages/component-viewer/component-viewer.scss
+++ b/src/app/pages/component-viewer/component-viewer.scss
@@ -1,20 +1,20 @@
-@import 'constants';
+@use '../../../styles/constants';
 
 app-component-viewer {
   font-weight: 400;
-  width: calc(100% - #{$sidenav-width + 1} - #{$content-padding-sidenav * 2});
+  width: calc(100% - #{constants.$sidenav-width + 1} - #{constants.$content-padding-sidenav * 2});
 
   // spacing for the component viewer
-  padding: 20px $content-padding-sidenav;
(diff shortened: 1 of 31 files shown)


src/app/pages/component-viewer/component-viewer.scss
```diff
@@ -1,20 +1,20 @@
-@import 'constants';
+@use '../../../styles/constants';
```
Splaktar, line 1: Is `@use` not compatible with? Or I guess, does `sass-loader` not support `stylePreprocessorOptions` with `@use`?

https://github.com/angular/material.angular.io/blob/8583f02ea14213d02a354a09cf82cbe0ab1d311a/angular.json#L73-L75

Splaktar (approved):
Love to see the return of normal `ng serve` and `ng build` times!

Hoping that `stylePreprocessorOptions` can still be supported.

src/app/pages/component-viewer/component-viewer.scss
crisbeto: I'm guessing it's a `sass-loader` issue, but I think that I prefer the relative path anyway. I found the shorthand confusing at first.

Splaktar: I just tested it out and it does look like `sass-loader` and WebStorm are both happy with using `@use` with `stylePreprocessorOptions`. I'd prefer to continue using these shorter paths if they don't cause some kind of build error.

PR #954, opened 2021-04-17
build: fix compilation errors in scenes

crisbeto:
The `yarn build:scenes` task was failing, because there were some `strictPropertyInitialization` errors.


Splaktar (approved):
Thank you for fixing this up!

PR #957, opened 2021-04-20
build: switch to eslint

crisbeto:
Switches the repo to use eslint instead of the deprecated tslint. Also adjusts the default config to match our previous setup.


.eslintrc.json
```diff
@@ -0,0 +1,108 @@
+{
+  "root": true,
+  "ignorePatterns": [
+    "projects/**/*",
```
devversion, line 4: Does this path exist?

devversion (approved):

.eslintrc.json
crisbeto: Doesn't look like it, I'll remove it. It was generated by the migration schematic.

PR #961, opened 2021-04-25
perf: example rendering performance improvements

crisbeto:
When generating an example in Stackblitz, we have to create a `form` element and submit it to a specific URL. The `form` has to be created eagerly, because some browsers will block us from submitting it if it is created asynchronously.

As a result of this setup, we fire off a lot of HTTP requests when an example is rendered which slows the page down. These changes make the following improvements which shave off more than a second of scripting time when transitioning from the "Overview" to "Examples". I've used the datepicker examples as a benchmark.

1. Runs the HTTP requests outside of the Angular zone so that we don't trigger change detections once each request is resolved.
2. Caches the file content so that the user doesn't have to load the same file multiple times.

I've also fixed that the copyright still said "2020".

For reference, here's what the flame chart looks like at the moment:
Before

And this is after these changes:
After

### DIFF
--- a/src/app/shared/stack-blitz/stack-blitz-button.ts
+++ b/src/app/shared/stack-blitz/stack-blitz-button.ts
@@ -8,7 +8,6 @@ import {StackBlitzWriter} from './stack-blitz-writer';
 @Component({
   selector: 'stack-blitz-button',
   templateUrl: './stack-blitz-button.html',
-  providers: [StackBlitzWriter],
 })
 export class StackBlitzButton {
   /**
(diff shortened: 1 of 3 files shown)


src/app/shared/stack-blitz/stack-blitz-writer.ts
```diff
@@ -140,26 +143,30 @@ export class StackBlitzWriter {
...
+        file, baseExamplePath, isTest)));
 
       // TODO(josephperrott): Prevent including assets to be manually checked.
       if (data.selectorName === 'icon-svg-example') {
-        this._readFile(form, data, 'assets/img/examples/thumbup-icon.svg', '', isTest, false);
+        fileReadPromises.push(this._loadAndAppendFile(form, data,
+          'assets/img/examples/thumbup-icon.svg', '', isTest, false));
       }
 
-      Promise.all(templateContents.concat(exampleContents)).then(() => {
```
crisbeto, line 159: Fun fact: this `Promise.all` didn't actually do anything, because the passed-in array was filled with `undefined` values.

Splaktar: 😬

src/app/shared/stack-blitz/stack-blitz-button.ts
```diff
@@ -18,28 +17,31 @@ export class StackBlitzButton {
    * StackBlitz not yet being ready for people with poor network connections or slow devices.
    */
   isDisabled = false;
-  stackBlitzForm: HTMLFormElement | undefined;
   exampleData: ExampleData | undefined;
 
-  @HostListener('mouseover') onMouseOver() {
-    this.isDisabled = !this.stackBlitzForm;
+  /**
+   * Form used to submit the data to Stackblitz.
```
Splaktar, line 23:
```suggestion
   * Form used to submit the data to StackBlitz.
```

src/app/shared/stack-blitz/stack-blitz-button.ts
```diff
@@ -18,28 +17,31 @@ export class StackBlitzButton {
...
+   * Important! it needs to be constructed ahead-of-time, because doing so on-demand
+   * will cause Firefox to block the submit as a popup, because it didn't happen within
+   * the same tick as the user interaction.
```
Splaktar, line 26:
```suggestion
   * Important! It needs to be constructed ahead-of-time, because doing so on-demand
   * would cause Firefox to block the submission as a popup. This is because it would't
   * happen within the same tick as the user interaction.
```

src/app/shared/stack-blitz/stack-blitz-writer.ts
```diff
@@ -1,14 +1,16 @@
...
+import {Observable} from 'rxjs';
+import {shareReplay, take} from 'rxjs/operators';
 
 import {materialVersion} from '../version/version';
 
 const STACKBLITZ_URL = 'https://run.stackblitz.com/api/angular/v1';
 
 const COPYRIGHT =
-  `Copyright 2020 Google LLC. All Rights Reserved.
+  `Copyright ${new Date().getFullYear()} Google LLC. All Rights Reserved.
```
Splaktar, line 13: 👏🏼

devversion (approved):

Splaktar (approved):
LGTM, nice work!

jelbourn (approved):
lgtm

PR #978, opened 2021-05-07
fix(sidenav): unable to scroll sidenav as expected

### LINKED ISSUE #857: docs-bug(navigation-links): Cannot see bottom components on nav until all scrolled down
https://material.angular.io/components/table/examples

Try to get to Tree from the left nav. You'll see that it requires scrolling down the entire page

Screen Shot 2020-08-14 at 9 30 54 AM

jljohnson001:
Fixes a bug in sidenav where user has to scroll to bottom of the page to reach bottom of navigation

Fixes #857

### DIFF
--- a/src/app/pages/component-sidenav/component-sidenav.scss
+++ b/src/app/pages/component-sidenav/component-sidenav.scss
@@ -33,7 +33,7 @@ app-component-sidenav {
 
   .docs-component-viewer-nav-content {
     width: constants.$sidenav-width;
-    height: 100vh;
+    height: 80vh;
     overflow: auto;
 
     &::-webkit-scrollbar {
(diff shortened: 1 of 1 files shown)


Splaktar (changes requested):
Screen Shot 2021-05-07 at 17 04 19

Thank you for looking into this! However, it doesn't quite get us what we need as you can see in the screenshot. There is a big space between the bottom of the menu and the bottom of the page. 

I'll look into this a little more and try to suggest some changes that will not have this issue.

src/app/pages/component-sidenav/component-sidenav.scss
```diff
@@ -33,7 +33,7 @@ app-component-sidenav {
 
   .docs-component-viewer-nav-content {
     width: constants.$sidenav-width;
-    height: 100vh;
+    height: 80vh;
```
Splaktar, line 36: It's actually the top nav that we're not accounting for, which is `56px`.
```suggestion
    height: calc(100vh - 56px);
```

Splaktar (commented):
This seems to improve things while avoiding the gap.

Splaktar (approved):

Issue #962, opened 2021-04-27
docs-infra: update old branches to point to latest versions.json file


besimgurbuz:
I also try to open the PR's for this issue

<div style="display: flex;">
<img src="https://user-images.githubusercontent.com/33575384/116261542-6bbaf880-a780-11eb-8b39-230349e09417.png">

<img src="https://user-images.githubusercontent.com/33575384/116260599-7de86700-a77f-11eb-9d85-5e10a56354f5.png">
</div>

besimgurbuz:
Issue resolved with  #964 #963

Splaktar:
I accepted those 2 PRs because they were reasonable and improved upon what we have now.

However, we probably want to look at switching these old branches to use an approach like v6 does:
https://github.com/angular/material.angular.io/commit/59825de50ab289ab1171aa31bdee8d3aabec42bc

For branches `8.x`, `9.x`, `10.2.x`, and `11.x`, we should change
```ts
const versionUrl = 'assets/versions.json';
```
to
```ts
const versionUrl = 'https://material.angular.io/assets/versions.json';
```

besimgurbuz:
I would like to implement the desired approach to these branches. If there is still a need. @Splaktar

Splaktar:
That would be much appreciated, thank you!

besimgurbuz:
I think we can close this issue now. @Splaktar

PR #979, opened 2021-05-07
fix(version-picker): restore the versions.json

Splaktar:
- we can't remove this until we release v12 and switch the main site to it
- add CSP connect-src rule to allow getting the versions.json file
- allow downloading the versions.json from other origins like v10.material.angular.io

Relates to #962


crisbeto (approved):
LGTM

PR #988, opened 2021-05-22
feat: add cookie popup

crisbeto:
Due to legal requirements we have to have a cookie disclaimer popup. It is always shown until the user agrees to it. After the user has agreed, the popup won't be shown on subsequent sessions.

Fixes https://github.com/angular/components/issues/22746.

**Note:** currently this doesn't have any accessibility treatment, because I'm not sure what the correct approach would be. I assume that we'd have to move focus into the cookie popup since it has more than one button, but its `role` is currently either `status` or `alert` which may not make sense to receive focus and to contain buttons.

Angular_Material_UI_component_library_-_Google_Chr_2021-05-22_08-44-27

### DIFF
--- a/src/app/shared/cookie-popup/cookie-popup.html
+++ b/src/app/shared/cookie-popup/cookie-popup.html
@@ -1,0 +1,13 @@
+<div class="popup" *ngIf="!hasAccepted">
+  This site uses cookies from Google to deliver its services and to analyze traffic.
+
+  <div class="buttons">
+    <a
+      href="https://policies.google.com/technologies/cookies"
+      mat-button
+      color="accent"
+      target="_blank"
+      rel="noopener">More details</a>
+    <button mat-button color="accent" (click)="accept()">Ok, Got it</button>
+  </div>
+</div>
--- a/src/_app-theme.scss
+++ b/src/_app-theme.scss
@@ -12,6 +12,7 @@
 @use './app/shared/footer/footer-theme';
 @use './app/shared/navbar/navbar-theme';
 @use './app/shared/table-of-contents/table-of-contents-theme';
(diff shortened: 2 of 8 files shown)


src/app/shared/cookie-popup/cookie-popup.html
```diff
@@ -0,0 +1,13 @@
+<ng-template #content>
+  This site uses cookies from Google to deliver its services and to analyze traffic.
+
+  <div class="buttons">
+    <a
+      href="https://policies.google.com/technologies/cookies"
```
RitikDua, line 6: How about providing policy link via input?
such as
```
<app-cookie-popup policy="https://policy.link" />
```

crisbeto: This is a one-off component so it's fine to hardcode the link.

crisbeto:
@jelbourn I've reworked it based on our discussion.

tools/audit-docs.js
```diff
@@ -35,7 +35,9 @@ const MIN_SCORES_PER_PAGE = [
       'performance': 28,
       'seo': 98,
       'best-practices': 100,
-      'accessibility': 100
+      // We lose a few points on accessibility, because there isn't quite enough contrast on the
+      // buttons in the cookie popup. This is a known issue with the Material Design theming system.
+      'accessibility': 98
```
jelbourn, line 40: Can we tweak the colors to avoid dinging this?

crisbeto: Done. It looks like I only had to darken the original color by 5%.

jelbourn (approved):
LGTM

Kerrick:
Any chance that we could get this added to the library as a component users of Angular Material could use?

PR #990, opened 2021-05-24
fix: use theme color for link inside the API reference

crisbeto:
* Fixes that links inside the API reference weren't themed, causing them to have a poor contrast ratio.
* Simplifies the Sass in the API theme.

Fixes https://github.com/angular/components/issues/14615.


mmalerba (approved):
LGTM

PR #995, opened 2021-05-25
docs(material/core): Add core API docs

mmalerba:
See demo at: https://material2-docs-dev.web.app/components/categories

I don't necessarily think this is the best way to organize the docs for `@angular/material/core`, but I think its better to at least show them for now so we don't leave important components and classes completely undocumented.

Fixes https://github.com/angular/components/issues/8790


jelbourn (commented):
lgtm

PR #999, opened 2021-05-26
feat: more flexible carousels and fix homepage alignment

crisbeto:
Currently the carousels on the front page don't align with any other elements on the page which looks odd. This seems to be due to a limitation where the item size has to be passed in to the carousel.

These changes reimplement the carousel, making it more flexible and allowing it to handle any size of items and any spacing. The new carousel is used on the frontpage to lay out the items in a way that aligns with the rest of the content.

Before:
before

After:
after

### DIFF
--- a/src/app/pages/homepage/_homepage-theme.scss
+++ b/src/app/pages/homepage/_homepage-theme.scss
@@ -57,21 +57,5 @@
       }
       outline: none;
     }
-
-    .docs-featured-components-carousel-item:hover {
-      img {
-        transform: scale(1.2);
-      }
-
-      box-shadow:inset 0 0 0 1.5px mat.get-color-from-palette($foreground, divider);
-    }
-
-    .docs-featured-components-carousel-item:focus {
-      img {
-        transform: scale(1.2);
-      }
-
-      box-shadow:inset 0 0 0 1.5px mat.get-color-from-palette($foreground, disabled);
-    }
   }
 }
--- a/src/app/pages/homepage/homepage.html
+++ b/src/app/pages/homepage/homepage.html
@@ -43,9 +43,16 @@
   <mat-divider></mat-divider>
 
   <div class="docs-homepage-featured-components docs-homepage-carousel-row">
-    <h2>Featured components</h2>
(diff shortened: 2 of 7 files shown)


jelbourn:
IMO the wider cards don't look as good

@angular/components-team thoughts?

crisbeto:
At the very least I think that we need to align the header to the carousel. Otherwise none of the main elements are aligned to each other.

twerske:
In general, I like taking up more space with images vs white space. The "minimalist spread out design" is definitely being sunset in favor of more color/content (see [The Year of The Blog](https://medium.com/@usonesinbetween/2017-the-year-of-the-blob-a3d899c9b071))
I also think aligning the header and text to the page looks more cohesive.

A few thoughts:
1. Is it possible to increase both the height AND the width? I agree with "IMO the wider cards don't look as good" but I think there's opportunity to create consistency with the 'High quality, versatile and frictionless" sections
Screen Shot 2021-06-07 at 1 21 24 PM

2. nit: looks like the first carousel is still inlayed a tiny bit so the margins aren't quite aligned:
119725471-43521700-be70-11eb-8a19-4293f4089c67

3. alt idea if we feel making the carousel's taller is increasing scroll depth would be to move the "view all =>" to condense the height then go in and widen:
Screen Shot 2021-06-07 at 1 27 52 PM

crisbeto:
I've reworked it based on Emma's feedback. What do you think?

Angular_Material_UI_component_library_-_Google_Chr_2021-06-09_20-33-03
Angular_Material_UI_component_library_-_Google_Chr_2021-06-09_20-33-15

jelbourn:
Overall I think it looks good, would be good to get @twerske's opinion 

The guides page does look off, though, is this something from this PR? It's not on the published site:
image

crisbeto:
The guides page was fixed in https://github.com/angular/material.angular.io/commit/5ce2c1ab46e8c1289bc0834538489345e44e451d, but I just hadn't rebased the carousel branch in a few days. I've rebased it now.

twerske (approved):
lgtm - approving based on the visual screenshots that the larger/better use of space and alignment of elements looks great!

Splaktar (approved):
LGTM

PR #1001, opened 2021-05-28
fix: cache fetched docs

crisbeto:
Currently we fetch a doc every time we need to display it which means that user navigating between two pages will have to request the same information multiple times.

These changes add a caching layer since the information isn't going to change.

### DIFF
--- a/src/app/shared/doc-viewer/doc-viewer.ts
+++ b/src/app/shared/doc-viewer/doc-viewer.ts
@@ -7,6 +7,7 @@ import {
   ComponentFactoryResolver,
   ElementRef,
   EventEmitter,
+  Injectable,
   Injector,
   Input,
   NgZone,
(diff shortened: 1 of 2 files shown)


src/app/shared/doc-viewer/doc-viewer.ts
```diff
@@ -15,18 +16,34 @@ import {
...
+
+  constructor(private _http: HttpClient) {}
+
+  fetchDocument(url: string): Observable<string> {
+    if (this._cache[url]) {
+      return this._cache[url];
+    }
+
+    const stream = this._http.get(url, {responseType: 'text'}).pipe(shareReplay(1));
+    return stream.pipe(tap(() => this._cache[url] = stream));
```
devversion, line 36: My expertise with RxJS is limited, but is there are a reason to cache the stream in the `tap`? Is this to avoid caching cases where an initial XHR request failed?

devversion (approved):
LGTM. Nice change!

src/app/shared/doc-viewer/doc-viewer.ts
crisbeto: Yeah, I was trying to avoid caching a failed request.

PR #1002, opened 2021-05-28
refactor: remove default form field appearance in Stackblitz

crisbeto:
Companion PR to https://github.com/angular/components/pull/22837. Removes the default form field appearance so that we get a consistent result between the docs site and Stackblitz.


Splaktar (commented):
I guess that this stayed in after the move to change the component's default to `fill` was backed out of v9.

I thought that would be going back in as part of v10, v11, or 12, but it looks like it hasn't happened. I posted a question about the associated issue in https://github.com/angular/components/pull/22837#issuecomment-858793156.

crisbeto:
Merging this in since the related changes in the docs were merged a while ago.

PR #1003, opened 2021-05-28
fix: align and clean up table design

crisbeto:
* Aligns the design between the tables inside markdown and the tables in the API docs. I chose to go with the design from the API docs, because the separation between cells is much better.
* Fixes the poor color contrast on the table border by using the correct theming variables.
* Removes the table box shadow and fixes up some spacings to make the tables look better.

Before:
before

After (note that this applies to tables in the API docs too):
after


jelbourn (approved):
lgtm

devversion (approved):

PR #1004, opened 2021-06-02
fix(material/api-theme): fix nav items hover/focus contrast

### LINKED ISSUE #759: Hover and focus indicators on header nav items have insufficient contrast
image

In the image above, Components is hovered with #6338b0 and CDK is focused with #5B33A2.  The background is #673AB7 which makes the contrast ratios 1.05:1 and 1.17:1 respectively.  WCAG 2.1 requires a contrast ratio of at least 3:1 for UI elements.  

This is for the default theme, but the other themes have similarly low contrast ratios.

kian23kpt:
Fix hover and focus indicators on header nav items with sufficient contrast

Fixes #759


kian23kpt:
Screenshot (15)
Screenshot (18)

Splaktar (commented):
Thank you for the screenshots, but what is hovered in those? It looks like it's just showing the selected?

What does it look like when an item has `focus`?

Ideally, hover and focus will be a slightly different shade than selected.

kian23kpt:
> Thank you for the screenshots, but what is hovered in those? It looks like it's just showing the selected?
> 
> What does it look like when an item has `focus`?
> 
> Ideally, hover and focus will be a slightly different shade than selected.

Screenshot from 2021-06-13 11-17-19
Screenshot from 2021-06-13 11-17-44
Screenshot from 2021-06-13 11-18-12
Screenshot from 2021-06-13 11-18-54

Guides  hovered in screenshots & Components selected.

Splaktar:
@kian23kpt you verified that the contrast in the hovered text is sufficient?

src/app/shared/navbar/_navbar-theme.scss
```diff
@@ -13,6 +13,15 @@
 
     .docs-navbar, .docs-navbar-header {
       background: mat.get-color-from-palette($primary);
+
+      & a:hover, & a:focus {
+        background: rgba(0, 0, 0, 0.4);
+        color: mat.get-color-from-palette($accent, 200);
```
Splaktar, line 19: I just want to make sure that `200` here isn't too dark to meet the contrast guidelines.

Splaktar (approved):
LGTM

kian23kpt:
> @kian23kpt you verified that the contrast in the hovered text is sufficient?

Yes, checked!

src/app/shared/navbar/_navbar-theme.scss
Splaktar: What was the purpose of using the accent color here instead of just inheriting the color? Was the shade of the background not enough to differentiate selected versus hovered/focus states? or was there a contrast issue?

kian23kpt: > What was the purpose of using the accent color here instead of just inheriting the color? Was the shade of the background not enough to differentiate selected versus hovered/focus states? or was there a contrast issue?

Used only to differentiate, similar to the sidebar. but by removing it, there is no problem of contrast.

PR #985, opened 2021-05-13
Fix 100% height in docs-guide

afnecors:
## Before 
image

## Fixed
image

PR #1007, opened 2021-06-06
fix: guides overlapping footer on narrow screens

crisbeto:
Fixes that the guide tiles can overlap the footer on smaller screens.

Fixes https://github.com/angular/components/issues/22860.

### DIFF
--- a/src/app/pages/guide-list/guide-list.scss
+++ b/src/app/pages/guide-list/guide-list.scss
@@ -1,6 +1,9 @@
 :host,
 .docs-guide {
-  height: 100%;
+  // Ensures that the footer is pushed to the bottom if the page is too short.
+  display: flex;
+  flex-direction: column;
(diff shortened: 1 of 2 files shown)


devversion (approved):
LGTM. tested locally.  Looks like a unit test fails though: 

```
Chrome Headless 90.0.4430.72 (Linux x86_64) GuideList should set the footer below the bottom of the given view FAILED
	Error: main content should take up the full given height: Expected 1103 to be 1200.
```

crisbeto:
I decided to remove the test. It just doesn't feel like something that we should be verifying in a unit test.

devversion:
agreed 👍

Splaktar:
There was already some work in progress for this that a contributor was helping with in https://github.com/angular/material.angular.io/pull/985.

PR #1010, opened 2021-06-07
fix: fix cookies dialog button contrast

### LINKED ISSUE #1008: New cookies dialog is not accessible in dark mode: contrast
# Purple and Green

Screen Shot 2021-06-06 at 20 02 17
Screen Shot 2021-06-06 at 20 02 40

# Pink and Blue-grey

Screen Shot 2021-06-06 at 20 03 06

kian23kpt:
Fix cookies dialog button contrast in dark mode

Fixes #1008


kian23kpt:
Screenshot (22)
Screenshot (21)

Now in dark mode have sufficient contrast but is not color accent type, 
and light mode have no change.
@Splaktar

src/app/shared/cookie-popup/cookie-popup.html
```diff
@@ -5,9 +5,8 @@
     <a
       href="https://policies.google.com/technologies/cookies"
       mat-button
-      color="accent"
       target="_blank"
       rel="noopener">More details</a>
-    <button mat-button color="accent" (click)="accept()">Ok, Got it</button>
+    <button mat-button (click)="accept()">Ok, Got it</button>
```
Splaktar, line 10: Can we use `color="primary"` here and above and not need the extra CSS? or does that use a hue that is too light in dark mode?

src/app/shared/cookie-popup/_cookie-popup-theme.scss
```diff
@@ -16,6 +16,10 @@
         map.get(mat.$dark-theme-foreground-palette, secondary-text)
       );
       background: if($is-dark-theme, map.get(mat.$grey-palette, 50), #252525);
```
Splaktar, line 18: It seems like this isn't working? the two colors are reversed? If it's a dark theme, it should use the black background, i.e. `#252525` and not the almost white `50` from the grey palette.

Maybe this would fix it without the need to change any accent/primary colors.

Splaktar (changes requested):

Splaktar:
Thank you for working on a fix for this so quickly!

src/app/shared/cookie-popup/cookie-popup.html
kian23kpt: `color="primary"` in light mode have insufficient contrast

Screenshot from 2021-06-07 22-43-34

src/app/shared/cookie-popup/_cookie-popup-theme.scss
kian23kpt: Yes! is reverse right now!
but i like your idea, so with `color=primary` & popup background same as the theme we will have:

Screenshot from 2021-06-07 22-49-14
Screenshot from 2021-06-07 22-49-27

looks good to you?
@Splaktar

Splaktar: @crisbeto This LGTM. WDYT?

I don't like the idea of popping up a blinding white background dialog in dark mode. But I wanted to check if you had intended that?

Splaktar (approved):
LGTM

src/app/shared/cookie-popup/_cookie-popup-theme.scss
crisbeto: I was mimicking the color scheme from the Material snack bar. The only difference is that I made the background ~5% darker to meet the accessibility requirements.

src/app/shared/cookie-popup/cookie-popup.html
```diff
@@ -5,9 +5,9 @@
     <a
       href="https://policies.google.com/technologies/cookies"
       mat-button
-      color="accent"
+      color="primary"
       target="_blank"
       rel="noopener">More details</a>
-    <button mat-button color="accent" (click)="accept()">Ok, Got it</button>
+    <button mat-button color="primary" (click)="accept()">Ok, Got it</button>
```
crisbeto, line 11: I'm not sure whether the spec allows more than one primary button next to each other.

Splaktar: I'm not familiar with that guidance. Do you have a link to that part of the spec?

crisbeto: I couldn't dig anything up from the spec, but my thinking was that you only have one primary action in a component. In this case the main action is accepting cookies which is more important compared to reading about cookies, because people likely know about them already.

src/app/shared/cookie-popup/cookie-popup.html
```diff
@@ -5,9 +5,9 @@
     <a
       href="https://policies.google.com/technologies/cookies"
       mat-button
-      color="accent"
+      color="primary"
```
Splaktar, line 8:
```suggestion
```
Can you please check whether removing the color here still passes the a11y contrast checks in light and dark mode?

Splaktar (changes requested):

src/app/shared/cookie-popup/cookie-popup.html
Splaktar: Got it. We'll try removing the coloring from the "More Details" button.

src/app/shared/cookie-popup/cookie-popup.html
kian23kpt: Looks Good!

Screenshot from 2021-06-13 10-33-13


Screenshot from 2021-06-13 10-32-42

Splaktar (approved):
LGTM

PR #1025, opened 2021-06-30
fix: add postinstall script back to make component examples work

amysorto:


Splaktar (approved):
Thanks! it was accidentally removed.

PR #1029, opened 2021-07-11
fix: angular material footer logo

### LINKED ISSUE #1027: footer: Angular Material logo is incorrectly formatted for IE11
When viewing the docs in IE11, the dimensions of the Angular Material logo are incorrect in the footer.

image

This can be fixed by adding `width: 40px` to `.docs-footer-version.angular-material-logo` in the CSS.

image

kian23kpt:
Fixes Angular Material footer logo for IE11

Fixes #1027

### DIFF
--- a/src/app/shared/footer/footer.scss
+++ b/src/app/shared/footer/footer.scss
@@ -25,6 +25,7 @@
   flex: 1;
 
   .angular-material-logo {
+    width: 40px;
     height: 40px;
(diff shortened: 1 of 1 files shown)


Splaktar (approved):
In Chrome, the width is evaluated to be `37.1875px`, but I tested with this fix's value of `40px` and it seems to be fine.

Thank you for this fix!

PR #1033, opened 2021-08-03
perf: avoid triggering change detection on scroll

crisbeto:
Fixes that the table of contents was triggering change detections while scrolling.


devversion (approved):
👍

PR #1044, opened 2021-08-20
bump to 12.2.2

wagnermaciel:


devversion (approved):
LGTM. Assuming `master` here should target the `latest` releases. For `next`, will there be a separate branch like `v13-next`?

wagnermaciel:
> LGTM. Assuming `master` here should target the `latest` releases. For `next`, will there be a separate branch like `v13-next`?

Yep, I'm following this up with a `13.0.0-next.1` update

PR #1050, opened 2021-08-23
fix: temporarily disable Ivy on Stackblitz

crisbeto:
Stackblitz has a race condition that an cause issues when running ngcc against our packages, resulting in errors.

These changes temporarily disable Ivy until we can start using the linker.


devversion (approved):

jelbourn (approved):
lgtm

Splaktar (commented):
LGTM

PR #1058, opened 2021-08-29
fix: ensure that menu items are always sorted

crisbeto:
De-duplicates some logic and adds programmatic sorting of the menu items so that we don't have to do it manually. Currently the "Component harnesses" menu item is out of order.

### DIFF
--- a/src/app/shared/documentation-items/documentation-items.ts
+++ b/src/app/shared/documentation-items/documentation-items.ts
@@ -546,25 +546,9 @@ const DOCS: { [key: string]: DocItem[] } = {
   // docs more granularly than directory-level (within a11y) (same for viewport).
 };
 
-for (const doc of DOCS[COMPONENTS]) {
-  doc.packageName = 'material';
-  doc.examples =
-    exampleNames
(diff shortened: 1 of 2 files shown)


src/app/shared/documentation-items/documentation-items.ts
```diff
@@ -584,3 +568,14 @@ export class DocumentationItems {
...
+
+function processDocs(packageName: string, docs: DocItem[]): DocItem[] {
+  for (const doc of docs) {
+    doc.packageName = packageName;
+    doc.examples = exampleNames.filter(key =>
+      key.match(RegExp(`^${doc.exampleSpecs.prefix}`)) &&
+      !doc.exampleSpecs.exclude?.some(excludeName => key.indexOf(excludeName) === 0));
+  }
+
+  return docs.sort((a, b) => a.name.localeCompare(b.name));
```
devversion, line 580: I'm not sure if it would make sense to specify `en` as a locale here as the whole page is english, but just mentioning it here in case.

devversion (approved):

devversion (approved):

PR #1065, opened 2021-09-11
fix: remove duplicated meta tag

USSDlover:
Remove duplicated meta tag for `theme-color` at line 48 and keep the tag at line 14

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -11,7 +11,6 @@
   <meta name="apple-mobile-web-app-capable" content="yes">
   <meta name="apple-mobile-web-app-status-bar-style" content="translucent">
   <meta name="mobile-web-app-capable" content="yes">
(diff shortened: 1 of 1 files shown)


USSDlover:
@googlebot I signed it!

Splaktar (changes requested):
Thank you for catching this and working on a fix!

This one should actually be kept and the first one should be removed.

USSDlover:
> Thank you for catching this and working on a fix!
> 
> This one should actually be kept and the first one should be removed.

Please, it is my honor!

The first meta tag has been removed and the second one added.

USSDlover:
> Thank you for catching this and working on a fix!
> 
> This one should actually be kept and the first one should be removed.

Splaktar (approved):
LGTM, thanks!

PR #1067, opened 2021-09-14
fix: github icon was not centered on mobile

rapzo:
The class was already there but not used.


rapzo:
Just to add some context, this is how it currently looks on mobile:
image

As said in the description, the css class that fixes it was already there but never added to the link's class list.

PR #1075, opened 2021-09-25
fix: use more lenient version when generating stackblitz

crisbeto:
Reworks the logic that stamps out the required version in Stackblitz to be more lenient so that we don't get weird errors between releases.

Also removes the shared `materialVersion` constant since it's redundant.


devversion (approved):

sonukapoor:
Does this get deployed right away after merging into master?

crisbeto:
No, somebody needs to deploy it.

PR #1079, opened 2021-09-29
fix: use more lenient version when generating stackblitz (#1075)

crisbeto:
Cherry-pick of #1075 for the 12.x branch.

Reworks the logic that stamps out the required version in Stackblitz to be more lenient so that we don't get weird errors between releases.

Also removes the shared `materialVersion` constant since it's redundant.


amysorto (approved):
lgtm

PR #1093, opened 2021-10-28
refactor: update stackblitz examples write to use webcontainers for v13 compatibility

devversion:
Updates to the Stackblitz example writer:

* Simplifies opening mechanism by relying on the small/tree-shakable
  `@stackblitz/sdk` package instead of manually constructing the form.

* Updates the template files to reflect the files generated by a new
  CLI project generated with CLI v13.rc.

* Removes the second boilerplate/template for running test harness
  examples. These will run as part of the same CLI boilerplate (allowing
  us to reduce the burden of maintaining a second boilerplate).

* Updates the StackBlitz base template from Angular to `node` which is
  bringing in the new _WebContainer_ feature from StackBlitz. We need to
  use this as v13 packages are currently not supported by StackBlitz due
  to the APF v13 changes.

  Also WebContainer is recommended to be used in our case due to the complexities
  involed in running/replicating the CLI, while WebContainers could simply run them.
  with the benfit of users being able to experience Angular directly in the browser, as
  they would locally.

  * Downside is that the Stackblitz examples will not work in Safari and
    Firefox yet. Stackblitz prints a good message explaining why.
  * For testing reproductions, the team needs to just download the zipped
    project and run it locally. This is another benefit of using the CLI
    directly in a WebContainer. It's a small trade-off for being able to
    actually rely on the CLI fully (mitigating potential cache issues
    like we have faced before).


jelbourn (approved):
lgtm

Splaktar:
> Removes the second boilerplate/template for running test harness
examples. These will run as part of the same CLI boilerplate (allowing
us to reduce the burden of maintaining a second boilerplate).

Hooray!

> Downside is that the Stackblitz examples will not work in Safari and Firefox yet. 

Doh! I can certainly see this being problematic for a small but passionate group of developers.

src/assets/stack-blitz/.gitignore
```diff
@@ -1,46 +1,2 @@
...
-npm-debug.log
-yarn-error.log
-testem.log
-/typings
-
-# System Files
-.DS_Store
-Thumbs.db
+.angular/
+node_modules/
```
Splaktar, line 2: Is this what Angular CLI v13 is generating for the `.gitignore` in a new project?

src/assets/stack-blitz/angular.json
```diff
@@ -3,84 +3,114 @@
...
   "newProjectRoot": "projects",
   "projects": {
-    "material-aio-example": {
-      "root": "",
-      "sourceRoot": "src",
+    "example-app": {
       "projectType": "application",
       "schematics": {
-        "@schematics/angular:component": {
-          "style": "scss"
```
Splaktar, line 12: Will the StackBlitz and CLI generate new components with `.css` files now?

src/assets/stack-blitz/package.json
```diff
@@ -1,40 +1,42 @@
...
+    "@angular/common": "13.0.0-rc.0",
+    "@angular/compiler": "13.0.0-rc.0",
+    "@angular/core": "13.0.0-rc.0",
+    "@angular/forms": "13.0.0-rc.0",
+    "@angular/material": "${version}",
+    "@angular/platform-browser": "13.0.0-rc.0",
+    "@angular/platform-browser-dynamic": "13.0.0-rc.0",
+    "@angular/router": "13.0.0-rc.0",
+    "rxjs": "~6.6.0",
+    "sass": "1.35.1",
```
Splaktar, line 24: I'm curious why this isn't in `devDependencies`?

src/assets/stack-blitz/package.json
```diff
@@ -1,40 +1,42 @@
...
-    "@angular/language-service": "^13.0.0-next.0",
-    "@angular/localize": "^13.0.0-next.0",
-    "@types/node": "^14.14.22",
-    "ts-node": "^8.10.2",
-    "typescript": "~4.3.5"
+    "@angular-devkit/build-angular": "13.0.0-rc.0",
+    "@angular/cli": "13.0.0-rc.0",
+    "@angular/compiler-cli": "13.0.0-rc.0",
+    "@types/jasmine": "3.10.1",
+    "@types/node": "^12.11.1",
```
Splaktar, line 33: Why is this moved back to v12?

src/assets/stack-blitz/src/theme.scss
```diff
@@ -0,0 +1,17 @@
...
+    color: (
+      primary: $candy-app-primary,
+      accent: $candy-app-accent,
+    ),
+    typography: mat.define-typography-config(),
+  )
+);
+
+@include mat.core($my-theme);
+@include mat.all-component-themes($my-theme);
```
Splaktar, line 17: 👍🏼  to using a custom theme.

src/assets/stack-blitz/src/index.html
```diff
@@ -1,13 +1,24 @@
 <!DOCTYPE html>
 <html lang="en">
-<head>
-  <meta charset="utf-8">
-  <meta name="viewport" content="width=device-width, initial-scale=1.0">
-  <link href="https://fonts.googleapis.com/icon?family=Material+Icons&display=block" rel="stylesheet">
-  <title>{{title}}</title>
-</head>
-<body class="mat-app-background">
```
Splaktar, line 9: Did the default background color change now?

src/assets/stack-blitz/.gitignore
devversion: yeah, this is the default generated output

src/assets/stack-blitz/src/index.html
devversion: probably will be different in dark mode. I'm re-adding it

src/assets/stack-blitz/package.json
devversion: this is what the CLI generates and I want to match that.

src/assets/stack-blitz/package.json
devversion: Not sure where this came from. I think it was a leftover from my old webcontainer VM. I'll re-generate the project with rc.2 where this can be removed completely.

src/assets/stack-blitz/.gitignore
devversion: actually not. I'll update to the gitignore the CLI generates. I added this manually because I didn't generate the project with git enabled.

PR #1094, opened 2021-10-30
refactor: update stackblitz CLI asset to match with CLI rc.2 output

devversion:
cc. @Splaktar


src/assets/stack-blitz/src/index.html
```diff
@@ -1,24 +1,17 @@
...
+  <meta charset="utf-8">
+  <title>${title}</title>
+  <base href="/">
+  <meta name="viewport" content="width=device-width, initial-scale=1">
+  <link rel="icon" type="image/x-icon" href="favicon.ico">
+  <link rel="preconnect" href="https://fonts.gstatic.com">
+  <link href="https://fonts.googleapis.com/css2?family=Roboto:wght@300;400;500&display=swap" rel="stylesheet">
+  <link href="https://fonts.googleapis.com/icon?family=Material+Icons" rel="stylesheet">
+</head>
+<body class="mat-typography mat-app-background">
```
devversion, line 13: I ran the `ng add` schematic on the boilerplate to match what people would have locally. Note that I also added `mat-app-background` here as the StackBlitz examples do not necessarily use the sidenav (and to make the background work for people tinkering with themes then)

src/assets/stack-blitz/src/styles.scss
```diff
@@ -1,11 +1,9 @@
 body {
-  font-family: Roboto, Arial, sans-serif;
+  font-family: Roboto, "Helvetica Neue", sans-serif;
   margin: 0;
   padding: 30px;
 }
 
-.version-info {
```
devversion, line 7: Also note that I removed this as it looks good without the extra CSS as well, and this matches more with the standard CLI project w/ `ng add @angular/material` then.

crisbeto (approved):
LGTM

PR #1095, opened 2021-11-01
feat: add support for reduced motion

crisbeto:
Makes it so that we disable all animations if the user prefers reduced motion.

### DIFF
--- a/src/app/shared/table-of-contents/_table-of-contents-theme.scss
+++ b/src/app/shared/table-of-contents/_table-of-contents-theme.scss
@@ -14,7 +14,6 @@
 
     .docs-link {
       color: mat.get-color-from-palette($foreground, secondary-text);
-      transition: color 100ms;
 
       &:hover,
       &.docs-active {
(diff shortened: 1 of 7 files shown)


src/app/shared/version-picker/version-picker.html
```diff
@@ -3,7 +3,7 @@
   {{materialVersion}}
   <mat-icon>arrow_drop_down</mat-icon>
 </button>
-<mat-menu #versionPicker="matMenu">
+<mat-menu #versionPicker="matMenu" xPosition="before">
```
crisbeto, line 6: Not necessary for the reduced motion fix, but I think that it looks better when the version picker opens in the same direction as the theme picker.

src/app/shared/table-of-contents/_table-of-contents-theme.scss
```diff
@@ -14,7 +14,6 @@
 
     .docs-link {
       color: mat.get-color-from-palette($foreground, secondary-text);
-      transition: color 100ms;
```
crisbeto, line 17: I got rid of this transition, because it was barely noticeable anyway.

devversion (approved):

PR #1096, opened 2021-11-02
refactor: update stackblitz boilerplate to setup test environment

devversion:
We removed the test environment calls from the harness example spec
files so that the standard CLI `test.ts` file can initialize the
test browser environment. This commit updates the SHA of the docs
content and re-adds the `test.ts` init environment logic.

https://github.com/angular/components/commit/88b718e8d947def1dc8c03050a4fb12fbe1e4ace

### DIFF
--- a/src/assets/stack-blitz/src/test.ts
+++ b/src/assets/stack-blitz/src/test.ts
@@ -1,6 +1,11 @@
 // This file is required by karma.conf.js and loads recursively all the .spec and framework files
 
 import 'zone.js/testing';
+import {getTestBed} from '@angular/core/testing';
+import {
+  BrowserDynamicTestingModule,
+  platformBrowserDynamicTesting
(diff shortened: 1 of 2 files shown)


src/assets/stack-blitz/src/test.ts
```diff
@@ -11,10 +16,10 @@ declare const require: {
 
 // First, initialize the Angular testing environment.
 // TODO: Re-enable this once example test specs no longer init the test environment manually.
```
josephperrott, line 18: Looks like we can remove this comment.

josephperrott (approved):
LGTM

src/assets/stack-blitz/src/test.ts
devversion: good catch 👍

josephperrott (approved):
LGTM

PR #1097, opened 2021-11-03
Update to latest RC versions

jelbourn:
Updates to the latest RC versions for v13. This also reverts a change
from #1004 which made the hover color for the top nav buttons use the
accent color, which is **wildly** off-spec.


wagnermaciel (approved):
lgtm

PR #645, opened 2019-10-06
feat: update to 9.0.0-next.x and enable Ivy rendering engine

Splaktar:
## PR Checklist

Please check that your PR fulfills the following requirements:
- [x] The commit message follows [our guidelines](https://github.com/angular/components/blob/master/CONTRIBUTING.md#commit)
- [x] Tests for the changes have been added or weren't appropriate

## What is the current behavior?
The app is based on Angular, CLI, and Material version `8.x` and Ivy is disabled.

## What is the new behavior?
The app is based on Angular, CLI, and Material version `9.0.0-next.x` and Ivy is enabled.

## Stats
Stats from production builds deployed to Firebase Hosting (CDN).
The times in `ms` were the best scores of approximately 10 page reloads with cache disabled in Chrome `Version 77.0.3865.90 (Official Build) (64-bit)`.

Note that due to the way that the app is architected (how it pulls in example modules from the `angular/material2-docs-content` repo), we don't currently do any lazy loading.

See PR diff for different versions of libraries between the before/after.

### Landing Page before Ivy
transferred: 819 KB
resources: 4.1 MB
Finish: 637 ms
DOMContentLoaded: 435 ms
Load: 565 ms

### Landing Page after Ivy
transferred: 686 KB ✅ **-130 KB**
resources: 2.5 MB ✅ **-1.6 MB**
Finish: 571 ms ✅ approx. -66 ms
DOMContentLoaded: 392 ms ✅ approx. -43 ms
Load: 498 ms ✅ approx. -67 ms

### Landing Page with Ivy and build-optimizer enabled
transferred: 607 KB ✅ **-79 KB**
resources: 2.3 MB ✅ -0.2 MB
Finish: 570 ms ✅ approx. -1 ms
DOMContentLoaded: 372 ms ✅ approx. -20 ms
Load: 498 ms ✅ approx. 0 ms

### Checkbox Overview before Ivy
transferred: 846 KB
resources: 4.1 MB
Finish: 822 ms
DOMContentLoaded: 470 ms
Load: 741 ms

### Checkbox Overview after Ivy
transferred: 713 KB ✅ **-133 KB**
resources: 2.5 MB ✅ **-1.6 MB**
Finish: 809 ms ✅ approx. -13 ms
DOMContentLoaded: 419 ms ✅ approx. -51 ms
Load: 736 ms ✅ approx. -5 ms

### Checkbox Overview with Ivy and build-optimizer enabled
transferred: 634 KB ✅ **-79 KB**
[...]

PR #1099, opened 2021-11-03
Always load versions from latest material.angular.io

jelbourn:
PR #645 erroneously changed this to a relative url. We want this to always use the latest material.angular.io so that we don't have to update old branches for new major versions.


josephperrott (approved):
LGTM

PR #1102, opened 2021-11-09
refactor: transfer favicon to stackBlitz

jeripeierSBB:
A `favicon.ico` is stored in `assets/stack-blitz/src` but not sent to stackBlitz. So despite `favicon.ico` in assets, it is currently not used anywhere.

Another solution would be to delete `favicon.ico`.


devversion:
Thanks!

PR #1105, opened 2021-11-19
fix: bump angular version to resolve query issue

crisbeto:
We're currently locked into version 13.0.0 which has a problem with queries using `forwardRef` which affects several places in the docs like the navigation between docs and API examples.

These changes update to 13.0.1 which contains https://github.com/angular/angular/pull/44113 in order to resolve the issue.


devversion (approved):
LGTM. Interestingly the PR from @petebacondarwin indeed also fixed the issue even with AOT linking.

**Before**:
test1

**After**
after

It looks like before the `forwardRef` was incorrectly preserved and moved into the closure for the content queries but before Pete's fix the forward ref was never resolved. interesting!

PR #1117, opened 2022-01-27
Fix forking and saving for webcontainer stackblitz examples

devversion:
Fixes the forking and saving functionality for WebContainer StackBlitz
examples. The saving/forking functionality did not work (no proper error though)
because we delivered a binary `.ico` file to StackBlitz.

To fix the examples, we will remove the `favicon.ico` file, even the goal
was to keep the examples as equivalent as possible to a `ng new`-created app.

It's surprising this worked in the past, but the StackBlitz SDK mentions that no binary files
are supported and this breaks the saving, according to my investigation. Still would be good
if StackBlitz would have a better error indication I guess (lots of people submitted issues on their end)


crisbeto (approved):
LGTM

crisbeto (approved):

PR #1123, opened 2022-02-17
Fix docs-content version mismatch for v13.x docs deployment

devversion:
The docs app currently deployed to `material.angular.io` is for Angular
Components v13, but relies on docs-content from `master` i.e. v14-next.

We should scope the docs-content to v13.2.x (the latest and last minor for
v13) for the v13.x docs deployment.

### DIFF
--- a/package.json
+++ b/package.json
@@ -39,7 +39,7 @@
     "@angular/cdk-experimental": "^13.2.3",
     "@angular/common": "^13.1.0",
     "@angular/compiler": "^13.1.0",
-    "@angular/components-examples": "angular/material2-docs-content#master",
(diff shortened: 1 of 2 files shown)


devversion:
Context: https://github.com/angular/components/pull/24377

crisbeto (approved):
LGTM, but I'm a little concerned that we'll forget to bump this for v14.

devversion:
@crisbeto this will only go into the `13.x` branch. master in the docs repo will continue using `master` docs-content, and in v14 this needs to be changed to the proper v14 release branches (this is a little unspecified; would love this to be automated)

PR #1126, opened 2022-02-17
build: setup test to ensure docs-content is correct for deployment/branch

devversion:
It is currently very easy to have a mismatch of docs-content and the
Angular CDK/Material version -> deployment target. e.g. v14 docs-content
might be used while we actually deploy the doc site for v13

To make this process more obvious, consistent and safe, this commit
introduces some sanity checking for deployment and merging. The checking
works as followed. We define three "intuitive" types of branches:

1. `master` which always will be `next.material.angular.io`. Deployed frequently.
2. `stable` which always will be `material.angular.io`. Deployed frequently.
3. v11, v12 etc. branches. These are archived versions. These are _rarely_ re-deployed.

```
|--------------------------|--------|------------------|
| Deployment Url           | Branch | Release Train    |
|--------------------------|--------|------------------|
| material.angular.io      | master | latest           |
| next.material.angular.io | next   | next             |
|--------------------------|--------|------------------|
```

We will mostly focus on (1) and (2). The new tooling will ensure as part of CI
and as last resort when the actual deploy happens, that the docs-content is matching
the release-train we are going to publish the docs-site for. 

The scripting will consult the Angular components repository through the Github API
to find the correct branch for the docs-content. This is necessary because consider we
are publishing for v13. It's not clear what publish branch is used for this version though. e.g.
is it `13.0.x`, `13.1.x` and so on? The shared dev-infra code can tell, based on the release tool.

As a concrete validation example: if the latest stable release-train in `angular/components` is
`13.2.x`, then we will enforce that the docs-content is also set to `13.2.x`. Same applies for such
archived majors (useful when re-deployed), and for the `next` release-train (e.g. `v14.0.0-next.X`)

As part of this process change: Whenever we cut a new major, stable can be branched-off and
updated to the new major. That should be similar to the old process. The validation scripts would
not need any manual updates, so it's easy to switch to a new major.

### DIFF
--- a/tools/docs-content/check-package-json-ci.js
+++ b/tools/docs-content/check-package-json-ci.js
@@ -1,0 +1,61 @@
+#!/usr/bin/env node
+
+/**
+ * @license
+ * Copyright Google LLC All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+
+const {checkPackageJsonForReleaseTrain} = require('./check-package-json');
+
+/**
+ * Below is a table outlining our strategy for validating the release train.
+ * The branch in `angular/material.angular.io` will be used as indicator
+ * for the intended deployment URL or release train:
+ *
+ * |--------------------------|--------|------------------|
+ * | Deployment Url           | Branch | Release Train    |
+ * |--------------------------|--------|------------------|
+ * | material.angular.io      | master | latest           |
+ * | next.material.angular.io | next   | next             |
+ * |--------------------------|--------|------------------|
+ *
+ * e.g. if a new commit is pushed for `next`, then we will check that the
+ * docs-content matches the branch of the `next` release train.
+ *
+ * For older, branched-out majors (like `12.x`), this check will not be run
+ * in case commits are pushed to such "archive" branches.
+ */
+function getReleaseTrainForCiBranch() {
+  switch (process.env.CIRCLE_BRANCH) {
+    case 'master':
+      return 'latest';
+    case 'next':
+      return 'next';
+    default:
+      return null;
+  }
+}
+
+/**
+ * Checks that the proper docs content revision/branch is configured for the
+ * current branch. The release train is determined based on the current CI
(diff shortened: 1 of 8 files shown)


tools/docs-content/check-package-json.js
```diff
@@ -0,0 +1,58 @@
...
+  }
+
+  console.info('Expected Angular CDK/Material version:', train.version.format());
+  console.info('Release train branch: angular/components#' + train.branchName);
+
+  assertDocsContentPackageJsonRevision(train.branchName);
+}
+
+/**
+ * Asserts tha the project `package.json` file has the docs-content configured
```
wagnermaciel, line 39: tha --> that

wagnermaciel (dismissed):
LGTM

tools/docs-content/check-package-json-ci.js
```diff
@@ -0,0 +1,61 @@
+#!/usr/bin/env node
+
+/**
+ * @license
+ * Copyright Google LLC All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
```
Splaktar, line 8: Do we need this is every script given that we have https://github.com/angular/material.angular.io/blob/master/LICENSE for the whole repo?

We should probably bump https://github.com/angular/material.angular.io/blob/master/LICENSE to 2022 as well.

devversion: It's probably not needed but shouldn't hurt. We are doing this in all dev-infra shared scripts and in FW as well.

Yeah, seems like it. That could be a separate change I guess

devversion:
> For deployment, are we sure that this script will always be correct? Should we have some way to --force or override it so that we can get a deploy out if the script has a bug in it?

The script _should_ always be correct. If it really fails incorrectly, then it's good to re-evaluate and see where the mistake comes from (e.g. if the branch named incorrectly, or is the docs-content really correct, etc). In the worst case, the check can be just commented out in the Bash script. Should be fine.

devversion:
Closing since we will auto-deploy and this would no longer be needed

PR #1135, opened 2022-03-22
build: fix circleci rebase to work with other branches than `master`

devversion:
CircleCI always rebases branches on top of `master` currently, breaking
CI in all other branches like `13.x`.

We should let the orb auto-discover the target branch..


crisbeto (approved):
LGTM

PR #1138, opened 2022-04-02
Avoid analytics prompt of CLI in stackblitz examples

devversion:
The prompt will prevent the examples from running automatically in
StackBlitz, now requiring user action to proceed with serving.

We should disable this as this setting is not persisted across
StackBlitz projects anyway..


josephperrott (approved):
LGTM

PR #1152, opened 2022-06-14
Use javascript to get copyright date rather than hardcoding it.

benborra:


benborra:
Someone else made the same change that was merged, so no reason to keep this PR open anymore 😢

PR #1153, opened 2022-06-17
Rework caching header rules in an attempt to fix caching issues

devversion:
e.g. StackBlitz assets should not be cached indefinitely as otherwise e.g. we might end up with Angular v13 and CDK/Material v14, breaking examples..


wagnermaciel (approved):
lgtm

PR #1156, opened 2022-06-23
fix: add material-moment-adapter to stackblitz dependencies

zarend:
Adds `@angular/material-moment-adapter` and `moment` to the package.json
on stackblitz examples. This fixes the "Cannot find module 'moment' or
its corresponding type declarations." in some examples.

Fixes https://github.com/angular/components/issues/25136

### DIFF
--- a/src/assets/stack-blitz/package-lock.json
+++ b/src/assets/stack-blitz/package-lock.json
@@ -9,15 +9,17 @@
       "version": "0.0.0",
       "dependencies": {
         "@angular/animations": "~14.0.0",
-        "@angular/cdk": "~14.0.0",
+        "@angular/cdk": "14.0.3",
         "@angular/common": "~14.0.0",
         "@angular/compiler": "~14.0.0",
         "@angular/core": "~14.0.0",
         "@angular/forms": "~14.0.0",
-        "@angular/material": "~14.0.0",
+        "@angular/material": "14.0.3",
+        "@angular/material-moment-adapter": "14.0.3",
         "@angular/platform-browser": "~14.0.0",
         "@angular/platform-browser-dynamic": "~14.0.0",
         "@angular/router": "~14.0.0",
+        "moment": "^2.18.1",
         "rxjs": "~7.4.0",
         "tslib": "^2.3.0",
         "zone.js": "~0.11.5"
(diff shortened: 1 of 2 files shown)


src/assets/stack-blitz/package.json
```diff
@@ -17,9 +17,11 @@
     "@angular/core": "~14.0.0",
     "@angular/forms": "~14.0.0",
     "@angular/material": "${version}",
+    "@angular/material-moment-adapter": "${version}",
     "@angular/platform-browser": "~14.0.0",
     "@angular/platform-browser-dynamic": "~14.0.0",
     "@angular/router": "~14.0.0",
+    "moment": "2.18.1",
```
crisbeto, line 24: Should we set this to `^2.18.1` to be a bit more flexible?

zarend: ohh yeah, that's right, I changed it to `^2.18.1`.

devversion:
can you update the lock file please so that our users wouldn't need the dependency tree to be resolved? (this is usually the slower operation compared to fetching concurrently)

devversion:
It's a little inconvenient, but currently you just temporarily replace the placeholder with the latest CDK/Material version, run npm install, and then re-add the placeholder.

src/assets/stack-blitz/package-lock.json
```diff
@@ -9,15 +9,17 @@
       "version": "0.0.0",
       "dependencies": {
         "@angular/animations": "~14.0.0",
-        "@angular/cdk": "~14.0.0",
+        "@angular/cdk": "14.0.3",
```
zarend, line 12: Is that supposed to be `~14.0.0`? I used 14.0.3, which is the latest.

zarend:
Thanks, I updated the lockfile

devversion (approved):

zarend:
Is material.angular.io supposed to automatically redeploy after merging this, or do I need to manually deploy?

devversion:
@zarend It will be auto-deployed when a new commit lands in the COMP repo

zarend:
Thanks, I have a trivial docs fix in the comp repo, so I'll go ahead and land that rn #25151

PR #1158, opened 2022-07-06
Set up Google Analytics 4 along with keeping legacy Universal Analytics 

devversion:
We currently use Universal Analytics. This is deprecated in favor of
Google Analytics 4 and UA will stop processing hits in October 2023.

This change intends to prepare us for this migration, and to already
pre-populate our GA4 property (there is no way to migrate existing data
/properties into a GA4 property -- a new one needs to be created).

This will help us minimize the data gap so that we can:

* Continue to look at the UA property with the full time span until
  October 2023
* Can start using the GA4 property long-term in the future, starting
  with data even before Universal Analytics stops processing new data.

We need to keep the existing `analytics.js` setup. Initially we have
considered using `gtag.js` for both the UA and GA4 properties, as it
supports that, but that doesn't work with our strict trusted types
enforcement because it results in multiple `gtag.js` scripts (specific
versions for UA or GA4) that recreate the same trusted type policies.
This causes runtime errors and breaks the setup.

Instead, with continued use of `analytics.js` we have the benefit of
a good separation of trusted types + events and configuration. There is
some problematic with translation of Universal Analytics Events to GA4,
or the other way around (even though we don't use custom events
currently)

We also do not need to send page views for our GA4 property because GA4
with gtag supports this automatically (respecting the history state --
using the `Enhanced measurement events` setting in the UI).
For our UA legacy instance we continue to dispatch events manually. This
logic can be removed in the future.

More details can be found here:
https://docs.google.com/document/d/1aK8u4ZlXbqQ2wMqmgSX7Ces8iLgamC13oCoG6VeBruA/edit?usp=sharing&resourcekey=0-EVe-Rhnme3bj_pkz2RcOmw.


src/app/shared/analytics/format-error.ts
```diff
@@ -0,0 +1,28 @@
...
+      // strip the message from the stack trace, if present
+      .replace(error.message + '\n', '')
+      // strip leading spaces
+      .replace(/^ +/gm, '')
+      // strip all leading "at " for each frame
+      .replace(/^at /gm, '')
+      // replace long urls with just the last segment: `filename:line:column`
+      .replace(/(?: \(|@)http.+\/([^/)]+)\)?(?:\n|$)/gm, '@$1\n')
+      // replace "eval code" in Edge
+      .replace(/ *\(eval code(:\d+:\d+)\)(?:\n|$)/gm, '@???$1\n');
```
devversion, line 22: FYI: This logic is taken from `angular.io`. Seems fine to do this, but not sure if strictly needed at this point..

crisbeto (approved):

devversion:
Merged and landed in `main` and `15.x`

PR #1167, opened 2022-08-29
build: add bazel build

jbedard:
The use of the async sass API currently seems to break when dealing with the bazel/rules_js symlinking of npm modules so the `NG_BUILD_LEGACY_SASS` flag has been set to use the old API. This was a recent change switching to the async API: https://github.com/angular/angular-cli/commit/308e3a017f876bfc727e68803bfbce11e9d3396e

Depends on:
* ~~https://github.com/angular/material.angular.io/pull/1173~~
* https://github.com/angular/material.angular.io/pull/1172


.circleci/config.yml
```diff
@@ -50,6 +54,15 @@ commands:
           paths:
             - dist
 
+  # setup_bazel_rbe:
```
josephperrott, line 57: Are we planning to do this still?

.npmrc
```diff
@@ -0,0 +1 @@
+strict-peer-dependencies=false
```
josephperrott, line 1: Why is this necessary, should we be using resolutions instead?

scenes/e2e/protractor.conf.js
```diff
@@ -4,17 +4,41 @@
...
+    // runfiles tree so that subprocesses spawned in a different working directory can still find them.
+    process.env.CHROME_BIN = path.resolve(process.env.CHROME_BIN);
+    process.env.CHROMEDRIVER_BIN = path.resolve(process.env.CHROMEDRIVER_BIN);
+}
 
 /**
  * @type { import("protractor").Config }
  */
-exports.config = {
+let config = {
```
josephperrott, line 21: const?

.npmrc
jbedard: I believe yarn defaults to false while pnpm defaults to true so it needs to be set to use rules_js.

If using resolutions isn't required then I'm not sure if it should be done in this PR?

.circleci/config.yml
jbedard: After talking offline... it's not done today so we're not doing it in bazel.

scenes/src/app/scenes/slider/slider-scene.ts
```diff
@@ -13,7 +13,7 @@ export class SliderScene implements AfterViewInit {
   @ViewChild('volume') volume!: MatSlider;
 
   ngAfterViewInit() {
-    this.volume.focus();
+    (this.volume as any).focus();
```
josephperrott, line 16: Is this as any necessary?  `MatSlider` should have a `focus` method.

josephperrott (approved):

package.json
```diff
@@ -96,5 +95,15 @@
     "shelljs": "^0.8.4",
     "ts-node": "^8.10.2",
     "typescript": "~4.8.2"
+  },
+  "pnpm": {
+    "packageExtensions": {
+      "@nrwl/devkit": {
```
devversion, line 101: Do you know where nrwl/devkit is coming from? That looks unexpected

scenes/karma.conf.js
```diff
@@ -25,8 +37,14 @@ module.exports = function (config) {
...
     autoWatch: true,
-    browsers: ['Chrome'],
+    browsers: [isBazel ? 'ChromeHeadlessNoSandbox' : 'ChromeHeadlessLocal'],
     singleRun: false,
+    customLaunchers: customLaunchers,
     restartOnFileChange: true
   });
+
+
+  if (process.env['TRAVIS']) {
```
devversion, line 47: I don't think we use Travis. Is this an artifact from some testing?

devversion (approved):
I also went through it. Looks great overall, just some minor nits/questions

scenes/karma.conf.js
jbedard: No I'm not sure, it already existed in all the other karma configs

.circleci/config.yml
jbedard: I think this was only required for e2e browser testing which is not on CI today so we're not doing it in bazel either.

.bazelrc
```diff
@@ -0,0 +1,50 @@
...
+build --experimental_allow_tags_propagation
+
+# Don't check if output files have been modified
+build --noexperimental_check_output_files
+
+# Ensure sandboxing is enabled even for exclusive tests
+test --incompatible_exclusive_test_sandboxed
+
+# Increase the timeout for all tests
+# TODO: only for necessary tests
```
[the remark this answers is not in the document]
jbedard: I assume it was only added for e2e tests so I'll try just deleting it.

.circleci/config.yml
devversion: but the plan is to enable remote execution for build and testing right? I might be misunderstanding

package.json
jbedard: Looks like it is no longer being used, see yarn.lock in https://github.com/angular/material.angular.io/commit/ac0acaebc46e3c29e09fa36af3d323ed049574b2#diff-51e4f558fae534656963876761c95b83b6ef5da5103c4adef6768219ed76c2deL3703

scenes/src/app/scenes/slider/slider-scene.ts
jbedard: Yeah I don't understand this. I don't actually see any `focus` method on the class or in the compiled .d.ts?

https://github.com/angular/components/blob/6011b7d7e789fd9ad0c724df20501c2758f750e2/src/material/slider/slider.ts#L95

gregmagolan:
@devversion With Bazel 6.0.0rc1 remote execution now works with rules_js. It could be done as a follow-up to this PR pretty easily.

WORKSPACE.bazel
```diff
@@ -0,0 +1,76 @@
...
+)
+
+load("@aspect_bazel_lib//lib:repositories.bzl", "aspect_bazel_lib_dependencies")
+
+aspect_bazel_lib_dependencies()
+
+http_archive(
+    name = "aspect_rules_js",
+    sha256 = "515277ae357e62f52e29e0bfb60b73d2d062b8d00d21351d31f37c5bb275d4f5",
+    strip_prefix = "rules_js-1.5.3",
```
gregmagolan, line 17: Some nice per improvements have landed in rules_js since this release. Best to update to the latest of rules_js and  aspect_bazel_lib before landing.

scenes/karma.conf.js
devversion: Interesting, seems like a legacy artifact. We can clean-up ourselves, but this config did not have this travis check before, so we shouldn't add it here?

scenes/src/app/scenes/slider/slider-scene.ts
devversion: I think it might be because of recent breaking changes to the slider for Angular v15. Basically a whole rework. Maybe add a TODO here and keep the cast for now?

devversion (approved):
LGTM

devversion:
@jbedard do you want to merge this into a separate branch? I think we cannot merge this into `main` for now till at least Angular v15 is out (which is expected this very soon)

scenes/src/app/scenes/slider/slider-scene.ts
jbedard: I've added the TODO

scenes/karma.conf.js
jbedard: I think I was aligning this with karma.conf.js (in the root), I don't recall why 😬

jbedard:
>@jbedard do you want to merge this into a separate branch? I think we cannot merge this into main for now till at least Angular v15 is out (which is expected this very soon)

We can merge to a separate branch if you'd like to merge it into `main` later. It won't have any CI keeping it green in the meantime? (not that it's green atm anyway 😬).

jbedard:
Note the first commit is also on its own in https://github.com/angular/material.angular.io/pull/1172 if you'd prefer merging that on its own. It is also broken for I think legit reasons unrelated to my change? See that PR...

devversion:
@jbedard Yeah, separate branch might be best for now to not risk the v15 docs. We can merge to main soon then:

* You can still see CI by creating a PR from the feature branch to `main`.
* Your commit `fix TS compiler error` seems to bring in accidentally some symlinks e.g. for `blaze-bin`?
* Regarding the failing audit job:  Seems green on `main` unless I miss something & we run it periodically as a cronjob from the `angular/components` repo IIRC.

jbedard:
I'm looking at the audit stuff more atm and will post again when I have an update, and will sort out any issues with the bazel files (sorry I'm randomly rebasing/pushing trying to get it green atm).

Shall I change this PR to create a new branch? I'm not sure the best way to do that...

devversion:
I've created a branch for the Bazel migration. Just let me know when I should merge this PR

jbedard:
Actually I may have fixed the audits after all this time. It seems as if the default `light-server` host is `localhost` on the cli and `0.0.0.0` with the js api, and I guess using an IP for the host really kills the: "Best Practices", "Performance", and "Progressive Web App" scores 🤔 

It's passing in https://github.com/angular/material.angular.io/pull/1172 if you want to merge it into `main` right now. I'm looking to see why it's still red under bazel here...

jbedard:
Ok it's green under bazel as well now. Sorry for the confusion with me thinking it was a legit failure, I guess `0.0.0.0` is bad ux 😬 

You can do the merge into the branch 👍

PR #1169, opened 2022-09-03
fix: add dynamic copyright

Phosphorus-M:
Update the copyright to change it dynamically using JS, so now don't need to do a deploy just to update the footer.


crisbeto (approved):
LGTM

PR #1170, opened 2022-09-13
build: update to Angular 15

crisbeto:
Updates to Angular 15 and TS 4.8.


devversion (approved):
looks like there is a failure

PR #1172, opened 2022-09-30
test: run audit tests from isolated js script

jbedard:
Moving as much logic as possible out of the package.json into the actual test script. This makes it easier to run in general, specifically so it can be run as a single script from bazel (https://github.com/angular/material.angular.io/pull/1167).


tools/audit-docs.js
```diff
@@ -55,18 +79,53 @@ function formatScores(scores) {
   return formattedScores;
 }
 
+
+// Launch the light-server to run tests again
+const bundle = process.argv[2];
+const delay = process.argv[3];
```
josephperrott, line 85: I would love to see us validate this being a number so that if its run locally that we let the user know the delay wasn't applied since `NaN` is treated as `0`

Additionally, we don't need to have the or check in the set timeout because its always going to be a number, defaulting to zero already.

```suggestion
const delay = Number(process.argv[3]);
if (isNaN(delay)) {
  console.error(`The provided value "${process.argv[3]}" should be a number, exiting`);
  process.exit(1);
}
```

tools/audit-docs.js
```diff
@@ -55,18 +79,53 @@ function formatScores(scores) {
...
+const bind = 'localhost';
```
josephperrott, line 86: Rather than binding to localhost, we should bind to `0.0.0.0` (which is pretty safe since this is a test script, not a production server of any kind.

josephperrott (approved):
Overall LGTM, just a couple thoughts.

jbedard:
Note the audit test failing atm seems to be a legit failure?

package.json
```diff
@@ -24,13 +24,8 @@
     "publish-beta": "bash ./tools/deploy.sh stable beta",
     "scenes:generate": "ng e2e --port 4201 scenes && yarn scenes:optimize",
     "scenes:optimize": "node tools/optimize-scene-screenshots",
-    "test:audit:a11y": "node tools/audit-docs-a11y",
-    "test:audit:a11y:localhost": "run-p --race \"~~light-server -s dist/material-angular-io -p 4200 --quiet\" \"test:audit:a11y http://localhost:4200\" --",
-    "test:audit:a11y:ci": "run-p --race \"~~light-server -s ../dist/material-angular-io -p 4200 --quiet\" \"test:audit:a11y http://localhost:4200\" --",
-    "test:audit": "node tools/audit-docs",
-    "test:audit:localhost": "run-p --race \"start:static\" \"test:audit http://localhost:4200 2000\" --",
```
jbedard, line 31: Note this was previously broken because there is no `start:static`. If i switch it to the same as the others using `dist/material-angular-io` then I get the same failure locally (before any changes from this PR) as on CI in this PR.

jbedard:
Was merged along with the bazel work in #1167

PR #1174, opened 2022-10-12
feat: switch to MDC components

crisbeto:
Switches the docs site to use the MDC components and fixes some visual issues that came up as a result.


crisbeto:
It looks like the test build error was already in `main` so it's probably not related to my changes.

devversion (approved):

PR #1191, opened 2023-03-05
build: update to aspect_bazel_lib 1.28.0 and aspect_rules_js 1.20.1

gregmagolan:
This is a pre-factor to hoisting this repository into the components repository. May as well get it onto the latest version of rules_js first.

@devversion, the latest version of rules_js will keep your pnpm-lock.yaml file up-to-date automatically (when your source-of-truth is a yarn.lock file) when any of the inputs to the npm_translate_lock repository rule change. The `.aspect/rules/external_repository_action_cache/npm_translate_lock_LTE4Nzc1MDcwNjU=` file acts like an action cache to know when to re-run `pnpm import` to update the pnpm lock file.


devversion (approved):
LGTM.

For later reference: This means we are maintaining two lock files. Something we will be able to fix long-term by switching to full pnpm.

gregmagolan:
I also recommend setting `ASPECT_RULES_JS_FROZEN_PNPM_LOCK` on CI so that if the pnpm lock file is out of date you'll get a red build. https://docs.aspect.build/rules/aspect_rules_js/docs/pnpm#update_pnpm_lock

PR #1198, opened 2023-04-12
build: account for APF v16 fesm2022 bundles

devversion:
The lazy loading of examples needs to be adjusted to generate chunks for all FESM2022 bundles. The FESM2020 folder no longer exists as of APF v16.


alan-agius4 (approved):
LGTM

PR #1200, opened 2023-04-15
feat: convert to standalone

crisbeto:
Converts the docs app to standalone. Most of the changes were made using the migration script, except for some code formatting updates and cleaning up a runtime error. There are still some modules that can be cleaned up, but they'll require more refactoring.


devversion (approved):
LGTM. Note that I did not validate all `imports` of standalone components manually. As long as tests works, and CI passes, seems fine to me

PR #1206, opened 2023-05-11
build: update to Angular 16 final

crisbeto:
Updates the site and the Stackblitz template to Angular 16 final.


gregmagolan:
@crisbeto Looks like after rebase the pnpm-lock file was populated with the problematic `registry.npmjs.org/` prefixes again which don't work with rules_js and transitive npm deps with aliases in them.

If you cherry pick this commit is should resolve it for you https://github.com/aspect-forks/material.angular.io/commit/95d71f3cdea501c82d0141a2761000b138354578

devversion:
I've merged your PR that combines Kristiyan's commit with the lock file re-generation.

PR #1208, opened 2023-05-13
chore: update to latest rules_js and regenerate lock file with pnpm 8

gregmagolan:
Possible work-around for https://github.com/aspect-build/rules_js/issues/1062 since the re-generated v6 pnpm lock file no longer has package names that start with `registry.npmjs.org/` in the lock file.

Re-generation was done with `pnpm import` with pnpm 8. The latest rules_js also uses pnpm 8 by default to run `pnpm import` when the `yarn.lock` or `package.json` change to update the lock file (as part of the `update_pnpm_lock = True` feature of npm_translate_lock)


gregmagolan:
cc @devversion

devversion (approved):

PR #1213, opened 2023-06-05
fix: replace use of turbo with yarn

Nemikolh:
This PR replace the use of `turbo` with `yarn` as `turbo` is being deprecated and StackBlitz has support for `yarn` v1 out of the box.

See related PR: https://github.com/angular/angular/pull/50576

### DIFF
--- a/src/app/shared/stack-blitz/stack-blitz-writer.ts
+++ b/src/app/shared/stack-blitz/stack-blitz-writer.ts
@@ -203,7 +203,7 @@ export class StackBlitzWriter {
         .replace(/material-docs-example/g, data.selectorName)
         .replace(/\${title}/g, data.description);
     } else if (fileName === '.stackblitzrc') {
-      fileContent = fileContent.replace(/\${startCommand}/, isTest ? 'turbo test' : 'turbo start');
+      fileContent = fileContent.replace(/\${startCommand}/, isTest ? 'yarn test' : 'yarn start');
     } else if (fileName === 'src/main.ts') {
       const mainComponentName = data.componentNames[0];
 
--- a/src/assets/stack-blitz/.stackblitzrc
+++ b/src/assets/stack-blitz/.stackblitzrc
@@ -1,7 +1,4 @@
 {
   "installDependencies": true,
-  "startCommand": "${startCommand}",
-  "env": {
-    "ENABLE_CJS_IMPORTS": true
-  }
+  "startCommand": "${startCommand}"
 }


src/app/shared/stack-blitz/stack-blitz-writer.ts
```diff
@@ -203,7 +203,7 @@ export class StackBlitzWriter {
         .replace(/material-docs-example/g, data.selectorName)
         .replace(/\${title}/g, data.description);
     } else if (fileName === '.stackblitzrc') {
-      fileContent = fileContent.replace(/\${startCommand}/, isTest ? 'turbo test' : 'turbo start');
+      fileContent = fileContent.replace(/\${startCommand}/, isTest ? 'npm test' : 'npm start');
```
crisbeto, line 206: We're using `yarn` in the examples, not `npm`.

Nemikolh: Oh thanks for pointing that out! That'll makes the PR much simpler then.

crisbeto (approved):
LGTM

src/app/shared/stack-blitz/stack-blitz-writer.ts
```diff
@@ -203,7 +203,7 @@ export class StackBlitzWriter {
         .replace(/material-docs-example/g, data.selectorName)
         .replace(/\${title}/g, data.description);
     } else if (fileName === '.stackblitzrc') {
-      fileContent = fileContent.replace(/\${startCommand}/, isTest ? 'turbo test' : 'turbo start');
+      fileContent = fileContent.replace(/\${startCommand}/, isTest ? 'yarn test' : 'yarn start');
```
devversion, line 206: I actually think we wanted `npm` right? The lock file is already a NPM lock file

crisbeto: I converted the Stackblitz template to yarn since npm was timing out. See https://github.com/angular/material.angular.io/pull/1194

devversion: Ah great! @d3lm what will be the Yarn support strategy with WebContainers? For angular.io we don't use Yarn

d3lm: @devversion How do you mean? For yarn we currently support yarn v1. Or maybe I am misunderstanding.

PR #1225, opened 2023-09-26
allow data uris as images

mmalerba:


firebase.json
```diff
@@ -45,7 +45,7 @@
           {
             "key": "Content-Security-Policy",
             // Keep in sync with `angular.json`.
```
mmalerba, line 47: I'm not sure what this comment means, I don't see this being set in angular.json

crisbeto (approved):

firebase.json
devversion: It's this one: https://github.com/angular/material.angular.io/blob/main/angular.json#L114

devversion (approved):
LGTM but we should update `angular.json` for similar dev behavior

firebase.json
mmalerba: Huh, don't know why I couldn't find that

PR #1238, opened 2024-02-14
Update to Node 20 and Angular 17.1

crisbeto:
Updates the build to Node 20 and Angular 17.1 to make the 17.2 release smoother.


WORKSPACE.bazel
```diff
@@ -26,7 +27,17 @@ load("@rules_nodejs//nodejs:repositories.bzl", "nodejs_register_toolchains")
 
 nodejs_register_toolchains(
     name = "nodejs",
-    node_version = "16.17.1",
+    # The below can be removed once @rules_nodejs/nodejs is updated to latest which contains https://github.com/bazelbuild/rules_nodejs/pull/3701
```
devversion, line 30: Isn't there a release for this?

devversion (approved):

WORKSPACE.bazel
crisbeto: I tried dropping the `node_repositories`, but it failed for find the Node version so I assume it hasn't been released/it doesn't have the exact version we need.

PR #1257, opened 2024-06-13
fix: remove elevation from nav bar

crisbeto:
The nav bar wasn't meant to have an elevation, it just wasn't working prior to a fix earlier today.

### DIFF
--- a/src/app/material-docs-app.ts
+++ b/src/app/material-docs-app.ts
@@ -10,7 +10,11 @@ import {CookiePopup} from './shared/cookie-popup/cookie-popup';
 
 @Component({
   selector: 'material-docs-app',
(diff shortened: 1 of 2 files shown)


src/app/material-docs-app.ts
```diff
@@ -10,7 +10,11 @@ import {CookiePopup} from './shared/cookie-popup/cookie-popup';
 
 @Component({
   selector: 'material-docs-app',
-  templateUrl: './material-docs-app.html',
+  template: `
```
crisbeto, line 13: I inlined these since it was overkill to have a separate file for 3 lines of HTML.

mmalerba (approved):

PR #1262, opened 2024-08-15
ci: convert bazel jobs to a PR Github Action workflow

clydin:
The Bazel variants of the existing CircleCI configuration have now been moved to a PR workflow for Github Actions. This is a first step to fully converting to the use of Github Actions within the repository. The non-Bazel jobs currently still executing with CircleCI both for PRs and main CI.


.github/workflows/pr.yml
```diff
@@ -0,0 +1,72 @@
...
+jobs:
+  lint:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Initialize environment
+        uses: angular/dev-infra/github-actions/npm/checkout-and-setup-node@db91da4e742cd081bfba01db2edc4e816018419b
+      - name: Install node modules
+        run: yarn install
+      - name: Execute Linting
+        run: yarn bazel test --test_tag_filters=+lint //...
```
josephperrott, line 30: Is there `+` supposed to be there on the filter?

I havent seen that syntax before.

josephperrott (approved):
LGTM

Just the one question of confirmation for the syntax on tag filtering

josephperrott (approved):

.github/workflows/pr.yml
clydin: Was there in the original command line but doesn't seem to be mentioned in the bazel docs.

Removed.

PR #1274, opened 2024-10-14
feat: show deprecated related information in `material/tooltip`

naaajii:
previously we weren't showing any other information other than `breaking-change` for deprecated fields, this commit adds a component that protrays information regarding deprecation in tooltips rather than `title` attribute

closes angular/components#29839

### DIFF
--- a/src/app/shared/doc-viewer/deprecated-tooltip.ts
+++ b/src/app/shared/doc-viewer/deprecated-tooltip.ts
@@ -1,0 +1,36 @@
+import {Component} from '@angular/core';
+import {MatTooltipModule} from '@angular/material/tooltip';
+
+/**
+ * This component is responsible for showing the 
+ * deprecated fields throughout API from material repo,
+ * 
+ * When deprecated docs content is generated like:
+ * 
+ * <div class="docs-api-class-deprecated-marker" 
+ *  title="Will be removed in v21.0.0 or later">
+ *  Deprecated
+ * </div>
+ * 
+ * It uses `title` attribute to show information regarding
+ * deprecation and other information regarding deprecation
+ * isnt shown either. 
+ * 
+ * We are gonna use this component to show deprecation 
+ * information using the `material/tooltip`, the information
+ * would contain when the field is being deprecated and what 
+ * are the alternatives to it which both are extracted from
+ * `breaking-change` and `deprecated`.
+ */
+@Component({
+  selector: 'deprecated-field',
+  template: `<div class="deprecated-content" 
+    [matTooltip]="message">
+  </div>`,
+  standalone: true,
+  imports: [MatTooltipModule],
+})
+export class DeprecatedFieldComponent {
+  /** Message regarding the deprecation  */
+  message = '';
+}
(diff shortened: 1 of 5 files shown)


naaajii:
image
image

naaajii:
(apologies for ping)

@crisbeto @devversion can we land this since angular/components#29873 is merged?

src/app/shared/doc-viewer/doc-viewer.ts
```diff
@@ -166,4 +170,30 @@ export class DocViewer implements OnDestroy {
     this._clearLiveExamples();
     this._documentFetchSubscription?.unsubscribe();
   }
+
+  _createTooltipsForDeprecated() {
+    // all of the deprecated markers end with `deprecated-marker` 
+    // in their class name
+    const deprecatedElements =
+      this._elementRef.nativeElement.querySelectorAll(`[class$=deprecated-marker]`);
```
devversion, line 178: Is the `$=` needed?

devversion (approved):
Ty. This looks great. A few minors nits, but otherwise that's great.

(Sorry for slow reponses; we've been on a full day summit)

src/app/shared/doc-viewer/doc-viewer.ts
naaajii: its shorter for like:

```js
querySelectorAll('.docs-api-deprecated-marker, .docs-api-class-deprecated-marker, .docs-api-class-interface-marker, .some-more')
```

src/app/shared/doc-viewer/deprecated-tooltip.ts
```diff
@@ -0,0 +1,36 @@
...
+  selector: 'deprecated-field',
+  template: `<div class="deprecated-content" 
+    [matTooltip]="message">
+  </div>`,
+  standalone: true,
+  imports: [MatTooltipModule],
+})
+export class DeprecatedFieldComponent {
+  /** Message regarding the deprecation  */
+  @Input() message!: string;
```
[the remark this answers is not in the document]
naaajii: i'm not sure how will i set the value for this input dynamically via ComponentPortal if we use signal based input.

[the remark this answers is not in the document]
devversion: Ohh. It doesn't need to be an input at all then, I think 😄

src/app/shared/doc-viewer/doc-viewer.ts
devversion: I see. How about adding a small comment illustrating the pattern + using `-deprecated-marker` as suffix test?

src/app/shared/doc-viewer/deprecated-tooltip.ts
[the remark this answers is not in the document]
naaajii: yes, true. I missed that! :smile:

src/app/shared/doc-viewer/doc-viewer.ts
naaajii: should be good now, lemme know if it needs rewording cause my english isn't so good.

crisbeto:
Looks like this got approved a while ago. Merging.

PR #1276, opened 2024-10-18
feat: delete lock file

wagnermaciel:
* The existence of a package-lock.json prevents npm i from installing the latest pre-release version


andrewseguin (approved):

crisbeto:
We should update the version, rather than deleting the lock file.

wagnermaciel:
How does this lock file get generated?

PR #1287, opened 2024-11-06
add system variables guide

andrewseguin:
Adds a new Guide that documents and demonstrates the available system-level CSS variables emitted in `mat.theme`, including its colors, typography, and elevation. Also documents the `mat.theme-overrides` API for overriding system-level theme properties.

This change includes a new feature for `DocViewer` so that it also accepts a `@Component`. Previously it always expected a URL to a markdown file to digest and render. Now guides can be Angular components in case more rich guides are needed.

### DIFF
--- a/src/app/pages/system-variables/system-variables.html
+++ b/src/app/pages/system-variables/system-variables.html
@@ -1,0 +1,242 @@
+<p>
+  Angular Material components depend on system variables defined as CSS variables through the
+  <span class="demo-surface-variable">material.theme</span>
+  Sass mixin. This page provides guidance and documentation for using these variables to
+  customize components.
+</p>
+
+<h2 id="colors">Colors</h2>
+
+<p>
+  Material Design uses color to create accessible, personal color schemes
+  that communicate your product's hierarchy, state, and brand. See Material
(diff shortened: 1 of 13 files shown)


src/app/pages/system-variables/system-variables.html
```diff
@@ -0,0 +1,242 @@
...
+<p>
+  The following colors are the most often used in Angular Material components. Use these
+  colors and follow their uses to add theme colors to your application's custom components.
+</p>
+
+<div class="demo-group">
+  <div class="demo-color-container">
+    <div class="demo-heading"
+         [style.background-color]="'var(--mat-sys-primary)'"
+         [style.color]="'var(--mat-sys-on-primary)'">
```
crisbeto, line 25: These `style` bindings aren't doing anything. You can either use a static attribute (`style="background-color: var(--mat-sys-primary)"`) or move them into the stylesheet. Same applies to the ones further down.

crisbeto (approved):

src/app/pages/system-variables/system-variables.html
andrewseguin: Ah yeah, I was planning to make this more programatic but it ended up getting over-engineered

amysorto (approved):

